Issue #39, opened 2023-03-28
Sentence case capitalisation for IEEE style


GeorgeHoneywood:
When generating a reference list for this bibtex, in IEEE style, hayagriva converts the `title` to be in sentence case:

```bibtex
@online{mdn-canvas-api,
  author  = {{MDN Contributors}},
  title   = {Canvas API},
  year    = 2022,
  url     = {https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API},
  urldate = {2022-11-02}
}
```

```shell
➜ hayagriva <file> reference --style ieee
MDN Contributors, “Canvas api.” https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API (accessed: Nov. 2, 2022).
```

This doesn't seem to be required by the [IEEE Citation Guidelines](https://ieee-dataport.org/sites/default/files/analysis/27/IEEE%20Citation%20Guidelines.pdf#page=3) or the [IEEE Reference Guide](https://ieeeauthorcenter.ieee.org/wp-content/uploads/IEEE-Reference-Guide.pdf) — is this behaviour intentional?

Unrelated: might be useful to have a generic numeric style that mimics the behaviour of biblatex's [default numeric style](https://mirror.apps.cam.ac.uk/pub/tex-archive/macros/latex/contrib/biblatex/doc/biblatex.pdf#page=75).

Jmcjack:
It looks like there is a bug in the `SentenceCase.apply()` method, as the `keep_all_uppercase_words` defaults to true for both the Title and Sentences cases. The behavior for the example in `lotr.bib` is different from your example. I added `LOTR` to the example and get the following:

`J. R. R. Tolkien, LOTR the Fellowship of the Ring LOTR, vol. 1, London: Allen & Unwin (AU), 1954.`
`MDN Contributors, “Canvas api.” https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API (accessed: Nov. 2, 2022).`

PR #44, opened 2023-04-26
Fix last word in title capitalization bug

Jmcjack:
Maintain consistency with last word in a bibtex entry title according to the `retain` flag.


Jmcjack:
I believe there are still bugs in the `SentenceCase.apply()` method re. Issue #39, but this is a minor (unrelated I think) fix.

PR #45, opened 2023-05-14
Add option to turn off abbreviation of journals

CMDJojo:
I think it would be great to have an option to not abbreviate journal names in the IEEE style. For our project, we would really need it. This PR adds that option, making sure that the default stays the same (i.e, to abbreviate journal names by default)


reknih:
Thank you!

PR #53, opened 2023-07-14
Fix spelling in doc comments

bluebear94:


reknih:
Thank you for verteiling these fixes :)

PR #63, opened 2023-09-25
Implement APA citation style

Zollerboy1:
Hey,

This PR adds an implementation of the APA citation style. I tried to comply with the APA guide as much as possible, but please point out mistakes to me if you find any.

This should partially fix typst/typst#1161 (as soon as this style is enabled there for the APA bibliography style).

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ to install and use Hayagriva on your terminal.
     - Author-Date references and citations
 - American Psychological Association (APA), 7th edition of the APA Publication Manual
     - References
+    - Author-Date citations, both parenthetical and narrative
 - Other in-text citation styles
     - Alphanumerical (e. g. "Rass97")
     - Author Title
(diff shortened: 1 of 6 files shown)


laurmaedje:
Thank you for this contribution! Unfortunately, we will probably not merging it because Hayagriva is currently being reworked to make use of [CSL](https://citationstyles.org/) citation styles, which will unlock a large library of citation styles without manual implementation in hayagriva. I will add an appropriate notice to the README.

Zollerboy1:
Oh okay.

Yeah, having this notice before would have been nice, as the README only said that you would be very happy to accept PRs for more citation and reference styles.

Is there an estimate for when the rework will be done? If it'll still be a while, perhaps we could already enjoy APA style citations in the meantime?

laurmaedje:
> as the README only said that you would be very happy to accept PRs for more citation and reference styles.

We're sorry about that, that was outdated.

> Is there an estimate for when the rework will be done? If it'll still be a while, perhaps we could already enjoy APA style citations in the meantime?

Within a few weeks probably. There will _probably_ not be a new Typst release in the meantime. Should it take longer than anticipated, I would be happy to merge this.

reknih:
Superseded by the CSL support in 0.9.0. Thank you anyways!

Issue #32, opened 2023-03-21
CSL?


bdarcus:
I'm admittedly biased, given I created it, but have you considered (also) supporting CSL citation styles and JSON input format?

I'm sure there are performance and other advantages to the rust-based styles, but there are thousands of CSL styles, as well as [citeproc-rs](https://github.com/zotero/citeproc-rs).

hcsch:
Perhaps TOML (or similar more simple languages, like JSON, as mentioned above) would also be a nice alternative to YAML for this use case, since YAML is known to be rather unintuitive/counter-intuitive in too many cases (see https://noyaml.com/ for example).

kmaasrud:
Using a well-defined format like CSL JSON, you'd easily be able to create a data structure that can be serialized/deserialized from any input format. This is possible already with the `citeproc-rs` crate. 

I agree with you @bdarcus, `hayagriva` really should leverage CSL, as the big contendors in bibliography management (notably Zotero) are fully in on it. 

@reknih, you should definitely hit up @cormacrelf and try to leverage his work wherever possible. The Rust ecosystem would definitely benefit from a proper citation processing library---I know that at least me and likely [the Tectonic guys](https://github.com/tectonic-typesetting) are interested!

bdarcus:
I dunno; I think toml would be a poor format for this purpose, and YAML is actually pretty good. Among other things, you can validate it's file using JSON schemas.

clbarnes:
citeproc-rs is possibly dead; readme calls it WIP, maintainer seems to have been active in the last couple of months but no commits to master since 2021.

kmaasrud:
> citeproc-rs is possibly dead; readme calls it WIP, maintainer seems to have been active in the last couple of months but no commits to master since 2021. 

Not very active, no, but it's still maintained, as evidenced by the PR opened 2 months ago that fixes a bug introduced by Rust 1.67. Also, it being part of the Zotero org, one would think it has some official weight.

That being said, Hayagriva is way more elegant IMO, and it actually exists on crates.io. Adapting to support CSL JSON shouldn't be that difficult (I am considering opening a PR for it.) CSL styles is another beast, though, but one that this library should definitely aim for supporting!

bdarcus:
[News](https://github.com/zotero/citeproc-rs/issues/61#issuecomment-1483791099) on `citeproc-rs`.

Basically they're stalled, with labor and technical hurdles, and need help to get the code in shape and released.

[...]

PR #66, opened 2023-10-19
CSL

reknih:
This PR implements CSL. cf. #32, typst/typst#2406, typst/typst#132
All known styles are tested to the extent that they don't cause a panic.

The following remains to do:
- ~Correctly map the CSL concept of [original metadata](https://docs.citationstyles.org/en/stable/specification.html#standard-variables) onto Hayagriva entries in Taxonomy.~
- ~Change Serial Number type to a dictionary and remove top-level `doi` etc. from `Entry`~
- `givenname-disambiguation-rule` is not applied to disambiguate names within a cite.
- `collapse="year"` deletes the year, but not the surrounding delimiters. It should rerender the item instead.
- `collapse="year-suffix"` and `collapse="year-suffix-ranged"` is ignored.
- `subsequent-author-substitute` is ignored.
- Decide whether to export members as-is or in submodules.

Maybe we should also use the [test suite](https://github.com/citation-style-language/test-suite), however, this would require injecting CSL variables directly, bypassing the `Entry` struct.


stellarpower:
Sorry if this is a silly question - should this allow me to use CSL YAML/JSON (output by Zotero) as the input format for the Hayagriva CLI? Or is this MR only for using the XML schema for the citation style?

yochem:
> Sorry if this is a silly question - should this allow me to use CSL YAML/JSON (output by Zotero) as the input format for the Hayagriva CLI? Or is this MR only for using the XML schema for the citation style?

I'm wondering this too. It seems that hayagriva is not able to parse CSL yaml files currently.

reknih:
Yes, this PR is only about CSL style files. I implemented CSL-JSON support for testing only (not exposed for library users), and right now, the structs are generic in such a way that CSL-JSON cannot be mixed with BibLaTeX/Hayagriva YAML. It could be a good contribution to allow users to bring CSL-JSON and mix it with other bibliographic entries.

PR #87, opened 2023-11-24
Fix Pulling Punctuattion into Quotation marks

### LINKED ISSUE #85: punctuation-in-quote seemingly not considered
I have a hayagriva entry like
```yml
JMLRefMan:
  type: report
  title: Java Modeling Language (JML) Reference Manual
  author:
    - Cok, David R.
    - Leavens, Gary T.
    - Ulbrich, Mattias
  date: 2022-07-01
  edition: 2
  url:
    value: https://www.openjml.org/documentation/JML_Reference_Manual.pdf
    date: 2023-11-23
```

And I use a CSL style with:
```xml
<locale xml:lang="en">
    <style-options punctuation-in-quote="true" />
    <!-- ... -->
  </locale>
```

Hayagriva gives me the ouput: 
```
[1] D.  R.  Cok,  G.  T.  Leavens,  and  M.  Ulbrich,  “Java  Modeling  Language  (JML)  Reference  Man-
ual”, 2nd ed., Jul. 2022. Accessed: Nov. 23, 2023. [Online]. Available: https://www.openjml.org/
documentation/JML_Reference_Manual.pdf
```

As can be seen, the punctuation is not moved into the title's quotation marks. Zotero handles this correctly, so the CSL seems to be correct.

I was also unable to find a test case for `punctuation-in-quote`.

Drodt:
Fix #85.

The issue is that in some instances (such as in #85), the quotes are not in a `Text` but another `Elem`.

My solution to this also considers `Text` if it is the only child of a surrounding `Elem`.

Since I don't know the architecture well, my hack might be supoptimal. Perhaps a `flatten` function on `ElementChild` makes more sense?

Also: How do you want me to add a test for this?

### DIFF
--- a/src/csl/mod.rs
+++ b/src/csl/mod.rs
@@ -2279,6 +2279,16 @@ impl<'a, T: EntryLike> Context<'a, T> {
                     .and_then(|p| p.0.last_mut())
                 {
                     Some(ElemChild::Text(f)) => &mut f.text,
+                    // Get the text element if it is contained in an `Elem`.
+                    Some(ElemChild::Elem(Elem { children, .. }))
+                        if children.0.len() == 1
+                            && matches!(children.0[0], ElemChild::Text(_)) =>
+                    {
+                        match &mut children.0[0] {
+                            ElemChild::Text(f) => &mut f.text,
+                            _ => unreachable!(),
+                        }
+                    }
                     _ => {
                         used_buf = true;
                         self.writing.buf.as_string_mut()


reknih:
Hey, thanks for the fix!
The testing setup is still a bit crap. We are using the citeproc.js test suite in (fittingly) the `citeproc.rs` file. You will need to run that file's tests manually with `--nocapture` to see how many tests passed before and after. We don't pass the full test suite at the moment so it will not fail CI if a regression occurs. I would like to add a file that contains all passing test cases so we can let CI fail if we got regressions.

PR #90, opened 2023-11-27
Update `strum` to eliminate `syn1`-dependency

CGMossa:
This is the end!

- [x] `comemo` has been updated to use `syn2`
- [x] `biblatex`-parser crate has been updated.
- [x] `biblatex` has to be _published_, and the new version should be added to `hayagriva` once that happens
- [x] this PR updates `strum` to latest version, that uses `syn2`. 
  
This will ensure that typst/typst doesn't have `syn1` in its dependency tree anywhere, and should thus yield compile time improvement.


CGMossa:
Latest version of `biblatex` is `0.9.1`, and this depends on `0.9`. I think that's fine, and thus I think this is ready to be merged.

Issue #94, opened 2023-12-01
Conference proceedings published as a book chapter


yochem:
I'm wondering how to cite a conference proceedings published as book chapter in APA:

Bedenel, A.-L., Jourdan, L., & Biernacki, C. (2019). Probability estimation by an adapted genetic algorithm in web insurance. In R. Battiti, M. Brunato, I. Kotsireas, & P. Pardalos (Eds.), Lecture notes in computer science: Vol. 11353. Learning and intelligent optimization (pp. 225–240). Springer. https://doi.org/10.1007/978-3-030-05348-2_21

I tried:
```yml
test:
  type: chapter
  author:
    - Bedenel, A. L.
    - Jourdan, L.
    - Biernacki, C.
  date: 2019
  title: Probability estimation by an adapted genetic algorithm in web insurance
  page-range: 225–240
  serial-number:
    doi: 10.1007/978-3-030-05348-2_21
  parent:
    type: book
    editor:
      - Battiti, R.
      - Brunato, M.
      - Kotsireas, I.
      - Pardalos, P.
    title: 'Lecture notes in computer science: Vol. 11353. Learning and intelligent optimization'
    publisher: Springer
```

Result:
Bedenel, A. L., Jourdan, L., & Biernacki, C. (2019). Probability estimation by an adapted genetic algorithm in web insurance. In Lecture notes in computer science: Vol. 11353. Learning and intelligent optimization (pp. 225–240). Springer. https://doi.org/10.1007/978-3-030-05348-2_21

When changing `editor` to `author`, we almost get the desired result, but missing the (Eds.) part of course:

Bedenel, A. L., Jourdan, L., & Biernacki, C. (2019). Probability estimation by an adapted genetic algorithm in web insurance. In R. Battiti, M. Brunato, I. Kotsireas, & P. Pardalos, Lecture notes in computer science: Vol. 11353. Learning and intelligent optimization (pp. 225–240). Springer. https://doi.org/10.1007/978-3-030-05348-2_21

It thus seems that `parent.editor` is skipped on book chapters, but `parent.author` is not.

As a side question: how to get the "Lecture notes in computer science: Vol. 11353. Learning and intelligent optimization" part right? I would expect to create a parent of the book with title "Learning and intelligent optimization" and volume 11353. That parent would be series/periodical with title "Lecture notes in computer science".

yochem:
[...]

PR #96, opened 2023-12-01
Fix editor reference (#94)

zepinglee:
The modify version of hayagriva tries to find the `editor` of the entry's container if it doesn't have `editor` itself. This is inspired by <https://github.com/typst/hayagriva/pull/82#discussion_r1397018367>.


reknih:
Thanks!

PR #99, opened 2023-12-05
Smart Quotation Marks in Formattable Strings.

### LINKED ISSUE #70: Smart Quotation Marks in Title
Assume an article has a title like `Applying "Design by Contract"`. If I just write it like this the quotation marks are neither inner quotation marks as expected nor are they cursive quotation marks. In other words, I would expect the output to be (in IEEE style): `“Applying ‘Design by Contract’,”`. 

I encountered the error in Typst, if that is relevant. Is there a way to let Typst format these strings?

Drodt:
Close #70.

I implemented a couple of changes to enable hayagriva to better render quotation marks in titles or other formattable strings.

The changes are:
When pushing a string to the buffer, we modify the string with the same heuristics as typst uses for smart quotes (taking into account any surrounding quotes created by hayagriva). To generate the quotation marks we add, we use the given style and language.

## Changes in passing tests

Adds 6 tests to passing but removes one: With the new changing of quotation marks, we change `flipflop_OrphanQuote` to a slightly different local test.

## Todo:
- [x] Better parsing of quotation marks (following typst's heuristics)
- [x] Additional Rust tests for quote funcions

## Todo in `biblatex`:
- [ ] Handling quotation commands (`\enquote`, `\mkbibquote`)

## Todo in `typst`:
- [ ] Move `SmartQuoter` to its own crate, or a util crate usable by typst and hayagriva


reknih:
Should this also work for single quotes? The `ChunkKind` could track the "doubleness" with a boolean.

DerDrodt:
> Should this also work for single quotes? The `ChunkKind` could track the "doubleness" with a boolean.

There is no need for that, and it might even produce problems.

1. When rendering, hayagriva automatically checks whether it should use inner or outer quotation marks.
2. Single quotes are more ambiguous in titles; think of `Dijkstra's algorithm` and the like.

So long as double quotes are part of the spec, I think it should be fine.

reknih:
If I'm reading your code correctly, BibTeX' normal chunks will continue to yield regular quotes. I think that we should parse BibTeX normal chunks to also emit the quote chunk so that both formats behave more similarly.

Drodt:
Another thing: I do think we will run into issues at some point by just parsing everything between `"` as a quote. (At some point someone will open an issue about weird behavior because they have two measurements in inches in their title.)

I think we should add support for simple typst commands like `quote`, `smallcaps`, etc. Thoughts?

reknih:
Ultimately, smart quotes are a heuristic. Typst implements it here: https://github.com/typst/typst/blob/c20b6ec6e157b753e90aa60ccd5b50f6f31cb7f6/crates/typst/src/text/smartquote.rs#L91. This handles something like quotes between numbers (e.g. for measurements in inches). It could be worthwhile to extract this into a crate and use the same code here and there. Additionally, I would propose to tokenize only the quote inside a chunk instead of the complete citation it wraps and approach this as purely string replacement instead of semantic.

Adding a non-standard chunk to `biblatex` is not very appealing to me. Maybe the Normal chunks could have a method that parses the appropriate commands and returns a vector with an enum?

Drodt:
Once you find the time, this PR is ready to merge @reknih @PgBiel

src/csl/quote.rs
```diff
@@ -0,0 +1,171 @@
+use citationberg::{taxonomy::OtherTerm, TermForm};
+
+use super::{taxonomy::EntryLike, Context};
```
PgBiel, line 3:
```suggestion
use citationberg::taxonomy::OtherTerm;
use citationberg::TermForm;

use super::taxonomy::EntryLike;
use super::Context;
```

src/csl/quote.rs
```diff
@@ -0,0 +1,171 @@
...
+
+/// A smart quote substitutor with zero lookahead.
+/// TODO: Move to its own crate.
```
PgBiel, line 6: I think this is fine as is. You can open a separate issue to discuss whether it'd make sense to use some shared impl between Typst and hayagriva, however.

```suggestion
```

src/csl/quote.rs
```diff
@@ -0,0 +1,171 @@
...
+            && (before.is_alphabetic() || before == '\u{FFFC}')
+        {
+            return "’";
+        }
+
+        // If the most recently opened quotation is of this kind and the
+        // previous char does not indicate a nested quotation, close it.
+        if opened == Some(double)
+            && !before.is_whitespace()
+            && !is_newline(before)
```
PgBiel, line 53: It appears that all newline characters in your function are whitespaces.[^1] [^2]

```suggestion
            && !before.is_whitespace()
```

[^1]: https://doc.rust-lang.org/std/primitive.char.html#method.is_whitespace
[^2]: `White_Space` in https://www.unicode.org/Public/UCD/latest/ucd/PropList.txt

src/csl/quote.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        // If we have a single smart quote, didn't recently open a single
+        // quotation, and are after an alphabetic char or an object (e.g. a
+        // math equation), interpret this as an apostrophe.
+        if !double
+            && opened != Some(false)
+            && (before.is_alphabetic() || before == '\u{FFFC}')
+        {
+            return "’";
+        }
```
PgBiel, line 47: I wonder if the object character check still applies here since they have a special meaning in Typst. But maybe it doesn't matter after all.

src/csl/quote.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+impl Default for SmartQuoter {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// Whether the character is an opening bracket, parenthesis, or brace.
+fn is_opening_bracket(c: char) -> bool {
```
PgBiel, line 93: Shouldn't matter in recent Rust, but for consistency:
```suggestion
#[inline]
fn is_opening_bracket(c: char) -> bool {
```

src/csl/quote.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// region as an all-uppercase ISO 3166-alpha2 code.
+    ///
+    /// Currently, the supported languages are: English, Czech, Danish, German,
+    /// Swiss / Liechtensteinian German, Estonian, Icelandic, Italian, Latin,
+    /// Lithuanian, Latvian, Slovak, Slovenian, Spanish, Bosnian, Finnish,
+    /// Swedish, French, Swiss French, Hungarian, Polish, Romanian, Japanese,
+    /// Traditional Chinese, Russian, Norwegian, Hebrew and Croatian.
+    ///
+    /// For unknown languages, the English quotes are used as fallback.
+    pub fn get<T: EntryLike>(ctx: &'s Context<'s, T>) -> Self {
```
PgBiel, line 135:
```suggestion
    /// Create a new `Quotes` struct with quotes taken from the current CSL locale's
    /// terms, falling back to `"` and `'` when not available.
    pub fn get<T: EntryLike>(ctx: &'s Context<'s, T>) -> Self {
```

src/csl/quote.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    /// The closing quote.
+    pub fn close(&self, double: bool) -> &'s str {
+        if double {
+            self.double_close
+        } else {
+            self.single_close
+        }
+    }
+}
```
PgBiel, line 171: We could have a `#[test]` here to test the Rust code independently from CSL too, maybe with a few trickier cases.

tests/local/flipflop_OrphanQuote.txt
```diff
@@ -0,0 +1,51 @@
+>>==== MODE ====>>
+citation
+<<==== MODE ====<<
+
+>>==== RESULT ====>>
+Nation of “Positive Obligations “ of State under the European Convention on Human Rights (1)
```
PgBiel, line 6: Interesting that the original test, which expected `"` at both positions, failed. I wonder if Hayagriva is just doing something that citeproc deliberately isn't? Can we find a reason for this to see if it is reasonable to override the test like that?

docs/file-format.md
```diff
@@ -430,6 +430,20 @@ publisher:
 Title and sentence case folding will always be deactivated if your item has set
 the `language` key to something other than English.
 
+Double quotation marks in a formattable string will be we transformed to
```
PgBiel, line 433: I do wonder if Typst should somehow wrap this in a smartquotes elements, which would render your changes potentially problematic... I'll try to ask the rest of the team on this.

PgBiel (commented):
Thanks! Overall, mostly fine. I can apply most of my suggestions myself, but I still have a few concerns which will need a bit of additional discussion.

src/csl/quote.rs
Drodt: I think it does not matter but I prefer to keep it as close to typst as possible.

tests/local/flipflop_OrphanQuote.txt
Drodt: On main the test passes. Hayagriva handles this correctly (aside from this branch using fancy quotation marks)

docs/file-format.md
PgBiel: After some initial discussion with part of the team, it's possible that we could want the smartquoter to emit an `ElemChild::SmartQuote` so that Typst can later emit a `smartquote` element, which would allow customization via show and set rules. The idea is then that the default plain text renderer would just convert it into text with the surrounding quotes, and the chosen quotes would be `None` if the CSL locale doesn't override the quote terms to allow for set rules to work on the Typst side.

However, we're still not sure regarding the algorithm. We would like to know how BibLaTeX approaches smart quoting so we can make sure existing `.bib` files behave at least somewhat as expected, converting one approach to another if necessary. It could be nice to have some further investigation in this regard.

Still, we won't make a decision before feedback from @reknih (who is currently unavailable), so don't worry about making too many changes right now, although feel free to investigate about BibLaTeX as that would be helpful.

Drodt: Sounds good. 

I already looked into how biblatex does this; I‘ll write it up some time next week

PR #102, opened 2023-12-12
Add abstract, annote, and genre fieldss

### LINKED ISSUE #101: No field for abstract in hayagriva format
There is no field that corresponds to bib(la)tex's abstract field. This is for example used by https://github.com/citation-style-language/styles/blob/master/apa-with-abstract.csl

Drodt:
This PR adds three fields, abstract, annote, and genre; all as `FormatableString`.

The first two are also extracted from the BibLaTeX input if present. The genre field, seemingly, has no equivalent in bib files.

This will close #101.

Let me know if I should change something or add more fields. Most other fields that are still missing seem more complex and require discussion how hayagriva should manage them.

### DIFF
--- a/src/csl/taxonomy.rs
+++ b/src/csl/taxonomy.rs
@@ -209,8 +209,13 @@ impl EntryLike for Entry {
     ) -> Option<Cow<'_, ChunkedString>> {
         let entry = self;
         match variable {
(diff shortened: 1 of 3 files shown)


reknih:
Thank you!

PR #105, opened 2023-12-13
Change publisher to dictionary

### LINKED ISSUE #34: publisher should be an object
Hey, I've just had a quick look at this project and it certainly looks interesting. I think using parent items instead of using `parent-title` etc. is a good choice. 
One thing to consider though: I suggest making publishers an object with attributes: name and location. Because there are rare cases where a book is being published by two publishers that each gave their own location. 
You will want to render this as:
Location A: Publisher A; Location B: Publisher B

We've had some discussions about this over at the CSL repo, and a similar change is planned, but it's much easier to make such a design decision when there's still not much technical debt...

DerDrodt:
Follows and fixes #34.

Changes:
- Changes the `publisher` field to be either a `FormatString` or a dictionary with `name` and `location`.
- `location` field remains **but only for events**; publisher location must be in the `publisher` field.
- **Update**: Publisher name is now optional

To discuss:
- Should the `location` field of an entry be renamed (e.g., `event-location`) or handled in another way?

TODO:
- Document changes to format.


MaxGietl:
Could this be implemented in a way that makes defining a standalone location for everything that isn't an event possible? In my university it seems to be standard practice to sometimes only display the location of a book, not the publisher itself, so I don't always have that information. Right now I have to define an empty string for the publisher, which obviously isn't ideal.
Currently this code is responsible for hiding the location if the publisher is None: https://github.com/typst/hayagriva/blob/d2db1fae5dcb258e649b1e78164c73320e95b664/src/csl/taxonomy.rs#L318, but obviously different in this new version.

DerDrodt:
> Could this be implemented in a way that makes defining a standalone location for everything that isn't an event possible?

That would kind of be the current implementation. Technically, in my implementation you can always define a location but
1. It is not intended as far as I am concerned;
2. Translating from bib(la)tex to hayagriva would ignore the location as you define it.
Cases like this is why this is a draft: We need some discussion on how the format should be changed.

> In my university it seems to be standard practice to sometimes only display the location of a book, not the publisher itself, so I don't always have that information.

If I may ask: Why? And what style are you using? It seems odd and is certainly not intended by most styles.

> Right now I have to define an empty string for the publisher, which obviously isn't ideal. Currently this code is responsible for hiding the location if the publisher is None:
> 
> https://github.com/typst/hayagriva/blob/d2db1fae5dcb258e649b1e78164c73320e95b664/src/csl/taxonomy.rs#L318
> but obviously different in this new version.

In my version, hayagriva behaves exactly the same: If no publisher is present, it shows no location.

If I am correct, and having a location but no publisher is not intended, adding an empty string seems fair.

MaxGietl:
This PR might have been the wrong place to bring this up because, as you say, it doesn't change the behaviour I'm talking about. I'm also not sure whether my initial impression is correct that it'd make exposing a location without a publisher more difficult.

I don't know which style they use, I'm not even sure they use a standard because at least one professor sometimes includes the publisher's name, sometimes they don't. I have briefly looked at the citations in two text books and one of them also only prints the location, so my professors don't seem to be that far from the standard of the field (social science). I will ask them though, if I get the chance.

Below are some example citations (in German, unfortunately):


What's the rationale behind hiding the location when the publisher isn't set? My workaround works well, and setting a single empty string isn't too bad, I was just very confused by this behaviour. But documenting that would resolve it as well.

DerDrodt:
> This PR might have been the wrong place to bring this up because, as you say, it doesn't change the behaviour I'm talking about. I'm also not sure whether my initial impression is correct that it'd make exposing a location without a publisher more difficult.

It’s supposed to make it clear that they belong together, usually. 

> I don't know which style they use, I'm not even sure they use a standard because at least one professor sometimes includes the publisher's name, sometimes they don't. I have briefly looked at the citations in two text books and one of them also only prints the location, so my professors don't seem to be that far from the standard of the field (social science). I will ask them though, if I get the chance.

That’s very weird. Line I said, it’s supposed to be the location where it’s published, which is odd without a publisher.

> Below are some example citations (in German, unfortunately): 

I’m German so that’s fine. If you can ask someone why they wrote it like that, that would be great!

> What's the rationale behind hiding the location when the publisher isn't set? My workaround works well, and setting a single empty string isn't too bad, I was just very confused by this behaviour. But documenting that would resolve it as well. Does bibtex do the same? I don't have any experience with that, unfortunately

BibTeX is too permissive here, it seems.

MaxGietl:
Okay after looking through some more random sociology and political science books, only two included the publisher's name and five didn't while all included the location. Obviously not representative, but I think it shows that that style is actually common place. And don't get me wrong: I think it's pretty weird too and it makes finding the correct book a lot harder but maybe hayagriva might want to support it because enough people seem to need it?

Drodt:
> Okay after looking through some more random sociology and political science books, only two included the publisher's name and five didn't while all included the location. Obviously not representative, but I think it shows that that style is at least somewhat common place.
> 
> And don't get me wrong: I think it's pretty weird too and it makes finding the correct book a lot harder but maybe hayagriva might want to support it because enough people seem to use it?

Strange... But yeah, in that case we should support it. The question is whether we consider it rare enough that setting the publisher's name to an empty string is the way to support it (in which case, we should document that), or just disregard this PR.

> Besides, what is the actual gain of forbidding it?

The gain of this PR is to structure the information in hayagriva better. And as far as I can see, in the vast majority of styles and for most tools, a location without a publisher is not intended.

> Also I think making the `name` field optional on the `Publisher` struct could be enough to support it? That should also make it clear that it's supposed to be the location of the publisher, even if no name is given

That's possible. But if someone just writes `publisher: some string`, should we still consider this the name?

MaxGietl:
> That's possible. But if someone just writes publisher: some string, should we still consider this the name?
I think that's a good compromise, maybe it could accept the following:
```yaml
guide:
   ...
   publisher: Megadodo

guide:
   ...
   publisher:
      name: Megadodo
      location: Ursa Minor

guide:
   ...
   publisher:
      location: Ursa Minor
```
Would that work?

DerDrodt:
That looks good

reknih:
I think that renaming `location` to `event-location` is too broad. What, for example, if you want to encode the Geo-Tag of a social media post or that you are citing the manuscript that your mentor wrote on Base Camp of Mount Everest?

src/lib.rs
```diff
@@ -505,9 +505,10 @@ entry! {
     #[serde(serialize_with = "serialize_one_or_many_opt")]
     #[serde(deserialize_with = "deserialize_one_or_many_opt")]
     "affiliated" => affiliated: Vec<PersonsWithRoles> | [PersonsWithRoles],
-    /// Publisher of the item.
-    "publisher" => publisher: FormatString,
-    /// Physical location at which the item was published or created.
+    /// Publisher of the item. May be a string or a dictionary with `name` and
+    /// `location` (both strings).
+    "publisher" => publisher: Publisher,
+    /// Physical location at the event took place.
```
PgBiel, line 511:
```suggestion
    /// Physical location at which the event took place.
```

src/lib.rs
```diff
@@ -505,9 +505,10 @@ entry! {
     #[serde(serialize_with = "serialize_one_or_many_opt")]
     #[serde(deserialize_with = "deserialize_one_or_many_opt")]
     "affiliated" => affiliated: Vec<PersonsWithRoles> | [PersonsWithRoles],
-    /// Publisher of the item.
-    "publisher" => publisher: FormatString,
-    /// Physical location at which the item was published or created.
+    /// Publisher of the item. May be a string or a dictionary with `name` and
+    /// `location` (both strings).
+    "publisher" => publisher: Publisher,
```
PgBiel, line 510: Not sure if this is the appropriate location to describe the type, maybe we can just have something like "Publisher of the item, which may have a name and a location."

src/types/mod.rs
```diff
@@ -411,6 +411,58 @@ impl Display for QualifiedUrl {
...
+    /// Physical location at which the item was published or created.
+    pub fn location(&self) -> Option<&FormatString> {
+        self.location.as_ref()
+    }
+}
+
+impl FromStr for Publisher {
+    type Err = ChunkedStrParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
PgBiel, line 461: Could consider adding a short doc comment here explaining that the parsed string will correspond to the publisher's name.

PgBiel (commented):
This looks nice! Here are some observations (paging @reknih for further thoughts).

PR #106, opened 2023-12-13
Try to render an author by forcing its ibid to Different before using the bibliography style

MaxGietl:
In a prose or author citation hayagriva tries to render the author with the style from the bibliography if the citation doesn't render the author itself. This leads to some awkward results if the CSL style doesn't display the author if the same citation was used directly before (instead just printing ibid. for example). This issue can be fixed by pretending that the previous citation doesn't exist, specifically by forcing `ctx.instance.cite_props.speculative.ibid` to be `IbidState::Different`.

Example using Typst:
```typst
$e$ is roughly $2.72$. @smith

Furthermore, #cite(<smith>, form: "prose") shows that $pi$ is roughly $3.14$

#bibliography("bibliography.yaml", title: "Literature", style: "deutsche-sprache")
```

with bibliography.yaml:
```yaml
smith:
  type: Book
  title: Special Numbers
  author: Smith, John
  date: 2010
```

Currently this renders to:
image

With my changes it renders to:
image

For completeness sake, this is what a standalone prose citation without a previous citation looks like in this style (this doesn't change):
image

### DIFF
--- a/src/csl/mod.rs
+++ b/src/csl/mod.rs
@@ -1197,35 +1197,52 @@ impl<'a> StyleContext<'a> {
                 ctx.set_special_form(Some(SpecialForm::VarOnly(author_var)));
                 do_regular(ctx);
             } else {
-                // Render name from bibliography.
-                ctx.set_special_form(Some(SpecialForm::VarOnly(author_var)));
-                let needs_synthesis = if let Some(bibliography) = &self.csl.bibliography {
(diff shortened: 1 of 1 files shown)


MaxGietl:
Btw this only looks like a fairly large change with 44 added lines because some code is now behind an additional if statement. The actual changes are <10 lines

reknih:
Thank you!

(I have to admit, it took me a while to grok what's going on here)

PR #116, opened 2023-12-15
Fix docs for DOI

### LINKED ISSUE #108: Deny unknown fields
Starting with hayagriva I was confused why some fields (eg. `translator`) that are variables in CSL wouldn't do anything when added to my entries. The answer of course is to RTFM (which I did eventually and that's where I learned about `affiliated`) but an error message telling my the field name was invalid would have spared my sanity quite a bit.

That's why I'm asking whether it would make sense to add [`#[serde(deny_unknown_fields)]`](https://serde.rs/container-attrs.html#deny_unknown_fields) to the `NakedEntry` struct that is used for deserialization. 

I realize this could make it more annoying when using a yaml file for a newer file with an old version of hayagriva, but I for one would like to know if some information is getting omitted. The other problem I can think of is people storing additional information in those files that isn't supposed to be read by hayagriva, is that something someone does?

### LINKED ISSUE #110: DOI not available in CSL
### Description

doi is referenced in the csl file but not printed.

image

section in main.csl
```
<text variable="DOI" prefix="doi:" suffix=", "/>
<text variable="URL"/>
```
refs.yaml
```
bartels2005:
  type: Article
  title: "Homer Gets a Tax Cut: Inequality and Public Policy in the American Mind"
  author: "Bartels, Larry M."
  date: 2005
  publisher: "Perspectives on Politics"
  doi: "10.1017/s1537592705050036"
  url:
    value: "https://doi.org/10.1017/s1537592705050036"
    date: "2023-12-06"
  page-range: 15-31
  volume: 3
  number: 1
  issn: "15375927, 15410986"
```

csl works with https://editor.citationstyles.org/visualEditor/

image

### Reproduction URL

https://typst.app/project/rYTX-2xTFEVZl8eKErcj8O

### Operating system

Linux

### Typst version

DerDrodt:
- ~Fail on unknown fields;~
- Fix doc comments; 
- Fix documentation for DOI (fix #110)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -355,7 +355,6 @@ macro_rules! entry {
                 struct MyVisitor;
 
                 #[derive(Deserialize)]
-                #[serde(deny_unknown_fields)]
                 struct NakedEntry {
                     #[serde(rename = "type")]
                     entry_type: Option<EntryType>,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -115,7 +115,8 @@ quantized-vortex:
     title: Structure of a Quantized Vortex in Boson Systems
     date: 1961-05
     page-range: 454-477
-    doi: 10.1007/BF02731494
+    serial-number:
+        doi: 10.1007/BF02731494
     parent:
         issue: 3
         volume: 20
@@ -354,6 +355,7 @@ macro_rules! entry {
                 struct MyVisitor;
 
                 #[derive(Deserialize)]
+                #[serde(deny_unknown_fields)]
                 struct NakedEntry {
                     #[serde(rename = "type")]
                     entry_type: Option<EntryType>,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -354,6 +355,7 @@ macro_rules! entry {
                 struct MyVisitor;
 
                 #[derive(Deserialize)]
+                #[serde(deny_unknown_fields)]
```
reknih, line 358: You proposed to warn instead in the original issue. I'm aware that we have no infrastructure for warnings at the moment, however, I do agree with you that we should try to stay forward-compatible and that denying parsing is too strict here. Would you be okay with removing this from the PR for now?

Drodt: Do you want just the doc fixes in this PR and the warnings in another one or should I create warnings in this PR as well?

About warnings: I can certainly do that but I might not find the most elegant solutions and will most likely have to change the API in some parts.

reknih: I would prefer separate PRs so we can land the docs changes sooner!

It would be great if you did. I think there are two possible designs: Returning a `Result<(T, NonCriticalDiags[]), CriticalDiags[]>` or returning a pointer to a tracer struct that contains the warnings (like Typst does: https://github.com/typst/typst/blob/c20b6ec6e157b753e90aa60ccd5b50f6f31cb7f6/crates/typst/src/lib.rs#L84).

reknih:
Thanks!

PR #123, opened 2024-01-04
Use workspace folder instead of CWD for file paths in test

frozolotl:
This PR fixes tests failing when they are not executed in the correct working directory.


reknih:
Thanks!

PR #125, opened 2024-01-09
Fix page range rendering

DerDrodt:
This PR (together with one for citationberg) fixes two citeproc tests (those for Chicago/Chicago16 page ranges).

### DIFF
--- a/src/csl/rendering/mod.rs
+++ b/src/csl/rendering/mod.rs
@@ -85,6 +85,11 @@ impl RenderCsl for citationberg::Text {
                 {
                     render_typed_num(num.as_ref(), NumberForm::default(), var, None, ctx);
                 }
+                NumberVariableResult::Regular(n)
+                    if matches!(var, NumberVariable::Page) =>
+                {
+                    ctx.push_str(&n.to_str().replace("-", "–"))
+                }
                 NumberVariableResult::Regular(n) => ctx.push_str(&n.to_str()),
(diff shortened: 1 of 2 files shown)


src/csl/rendering/mod.rs
```diff
@@ -85,6 +85,11 @@ impl RenderCsl for citationberg::Text {
                 {
                     render_typed_num(num.as_ref(), NumberForm::default(), var, None, ctx);
                 }
+                NumberVariableResult::Regular(n)
+                    if matches!(var, NumberVariable::Page) =>
+                {
+                    ctx.push_str(&n.to_str().replace("-", "–"))
```
reknih, line 91: Does the CSL standard hard-code the en-dash here?

DerDrodt: Not really. Honestly, this is a quick fix so we can handle these cases correctly (for the most part). I will work rework the page ranges in general. They should be able to handle more than numbers. I should get around to that some time next week.

reknih: Got it

reknih:
Thanks!

By the way, are you and @Drodt the same person?

DerDrodt:
> Thanks!
> 
> By the way, are you and @Drodt the same person?

You're welcome! And yes. @Drodt is my work account and when I commit stuff while traveling it comes from my work laptop. Sometimes I forget to swap. I hope it's not too confusing!

PR #131, opened 2024-01-26
Update archive and passing tests

DerDrodt:
Update archive files (since the CSL attribute for givenname-rule was wrong) and the passing tests.

@reknih Why is `page_NoOption` passing for you? It should not! Expanded is the default in Hayagriva if no PageRangeFormat is given and that should produce a different output.


reknih:
It passed with the data and the `citationberg` version at 2e098f555ed06ad56ed62f56e33c7202f4cea0a5 for me. Now it does not anymore!

reknih:
Isn't the update of `src/csl/archive.rs` missing?

reknih:
I updated the archive in eb13ecfcdbd35b986650fe897bfd877d72a6367c and 22f4439379440e3200d024c0d7f8a9d5e48157df so I'm closing this for the time being.

PR #133, opened 2024-01-26
Fix Issue #122

### LINKED ISSUE #122: CSL: Position "ibid" isn't tested well.
In my csl file I have the following choose:
```xml
  <macro name="display_full">
      <choose>
        <if position="ibid"><text term="ibid."/></if>
        <else-if position="subsequent"><text macro="short"/></else-if>
        <else-if type="webpage"><text macro="webpage"/></else-if>
        <else-if type="book"><text macro="book"/></else-if>
      </choose>
  </macro>
  ```
  
  But when I cite the exact same source with the same locator, the subsequent position is shown, instead of `ibid` or `ebd.` in my case (german):
  
image

According to the specification, the position should be "ibid" for the cite in footnote 18:
> Preceding cite does have a locator: if the current cite has the same locator, the position of the current cite is “ibid”. If the locator differs the position is “ibid-with-locator”. If the current cite lacks a locator its only position is “subsequent”.

https://docs.citationstyles.org/en/stable/specification.html#choose

DerDrodt:
Changed `BibliographyDriver::finish` to not _take_ the locator from the speculative items. That lead to the locator of `last` to always be `None`, which caused #122.

Does anyone know why the `take` was there?

Fixes #122.

### DIFF
--- a/src/csl/mod.rs
+++ b/src/csl/mod.rs
@@ -347,8 +347,7 @@ impl<'a, T: EntryLike + Hash + PartialEq + Eq + Debug> BibliographyDriver<'a, T>
                 );
 
                 //     - Add final locator
-                res[i].items[j].cite_props.speculative.locator =
-                    mem::take(&mut res[i].items[j].locator);
+                res[i].items[j].cite_props.speculative.locator = res[i].items[j].locator;
             }
         }
 


reknih:
Likely premature optimization

PR #135, opened 2024-02-03
Wrong translation for "Online" pt-PT

chicoferreira:
The translation found in https://github.com/typst/hayagriva/blob/main/archive/locales/pt-PT.cbor for "Online" is "Em linha" which is wrong. "Em linha" is the literal translation for "On line", but there is no translation for many foreign words, including "Online".

Not sure if this is the correct way to edit the .cbor files. Waiting on your feedback.


reknih:
The CBOR files are generated from https://github.com/citation-style-language/locales. If we were to apply a change here, it would be deleted when we next update the files from Upstream. Best to file an issue there!

chicoferreira:
Hey @reknih. The fix has been merged in https://github.com/citation-style-language/locales/pull/303. Please update the locales in the next update or guide me how to do it. 

Thank you.

chicoferreira:
Okay found out that running the tests generates the locales. Will follow up this with a new PR.

PR #143, opened 2024-03-19
Update archive locales and styles

chicoferreira:
Follow up to https://github.com/typst/hayagriva/pull/135

This PR aims to update the archive locales and styles to the latest upstream of https://github.com/citation-style-language/locales, which has the "Online" translation for pt-PT fixed as discussed in the mentioned PR.


reknih:
Thank you for your upstream contribution! Since there is breakage in the APA test through the update, I instead updated just the locales in 22f4439379440e3200d024c0d7f8a9d5e48157df.

PR #144, opened 2024-03-20
Fix citation suppresses title when no author provided

### LINKED ISSUE #139: APA 7 - No title when referencing a book or article without author
I have an issue where the title of a source is suppressed when there is no author to the work. If I add a mock author it works as expected. This is the incorrect format that gets rendered now: 

```
In bibliography: (Date). Source.
In text: (year)

```
This is how I expect it to be formated:

```
In bibliography: Title. (Date). Source.
In text: (Title, year)
```

I am using the latest version of typst and this is the entry in the bibliography.yml file: 

```YML
ethics:
  type: Article
  title: "Good Research Practice"
  date: 2017
  publisher: Swedish Research Council
  url: https://www.vr.se/english/analysis/reports/our-reports/2017-08-31-good-research-practice.html
```

Here are the official APA guidelines for referencing sources with missing information: https://apastyle.apa.org/style-grammar-guidelines/references/missing-information

AstrickHarren:
From https://github.com/typst/citationberg/issues/4#issuecomment-2009709750
> I found the problem is due to `substitute` and `choose` working together. (Its actually a problem of `typst/hayagriva`)
> 
>    1. `substitute` will cause any subsequent `query` to variables, in this case, the `title`, suppressed if that variable is queried once before.
> 
>    2. `choose` actually calls `query` to `title`  two times, the first time to find out which branch makes sense and the second to render.
> 
>    3. therefore the second time is suppressed and resulting in no title.
> 
> 
> A simple fix can be to do a silent lookup on the first `query` of `choose`, much like what is done in `Test::will_have_info(..)`.
> 
> This is the output from OP's bib after applying the fix:
> 
> ```
> Definition and objectives of systems development. (2016, January 19). https://www.opentextbooks.org.hk/ditatopic/25323
> ```
> 
> However, the `journal` part does not seem to be there, which I think is not there even with author specified.

Fixes typst/citationberg#4
Fixes #139

### DIFF
--- a/src/csl/rendering/mod.rs
+++ b/src/csl/rendering/mod.rs
@@ -774,16 +774,17 @@ fn choose_children<F, R, T: EntryLike>(
 where
     F: FnMut(&[LayoutRenderingElement], &mut Context<T>) -> R,
 {
-    for branch in choose.branches() {
-        if branch.match_.test(BranchConditionIter::from_branch(branch, ctx)) {
-            return Some(f(&branch.children, ctx));
-        }
-    }
+    let supressed = ctx.writing.suppress_queried_variables;
+    ctx.writing.stop_suppressing_queried_variables();
+    let branch = choose
+        .branches()
+        .find(|branch| branch.match_.test(BranchConditionIter::from_branch(branch, ctx)));
+    ctx.writing.suppress_queried_variables = supressed;
 
-    choose
-        .otherwise
-        .as_ref()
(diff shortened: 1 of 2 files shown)


AstrickHarren:
As tests seems to build up, I think I can help with reorganizing them into functions/macros as well :)

reknih:
Thank you!

> As tests seems to build up, I think I can help with reorganizing them into functions/macros as well :)

That would be fantastic! I also aim to increase compliance with the `citeproc` test suite.

PR #145, opened 2024-03-21
Fix #142

### LINKED ISSUE #142: Additional page gets displayed when only one is specified in the bibliography entry
Since #104, I believe, single page numbers in the `pages` field (`.bib`) or in the `page-range` field (`.yml`) no longer render correctly. It always adds the following page as well, turning `p. 103`, for example, into `p. 103 -- 104`. Note that, whatever the CSL style uses as a word for `page` does not get pluralized.

This can be reproduced with just about any simple entry and many styles, like the default IEEE.
https://typst.app/project/rxfSNvq9A61PlAievgf9sa

DerDrodt:
Fix #142.

Page ranges are now `RangeInclusive<i32>`, which fits the semantics of a page range better. Also changes how single pages are turned to ranges: `4` now becomes the inclusive range `4..=4`, which fits.

Also, in citationberg, ranges where start and end are the same only render start.

Depends on typst/citationberg#12.

This is just a quick fix until I find time to rework page ranges.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ archive = ["ciborium"]
 csl-json = ["citationberg/json"]
 
 [dependencies]
-citationberg = "0.3.0"
+citationberg = { git = "https://github.com/typst/citationberg.git" }
 indexmap = { version = "2.0.2", features = ["serde"] }
 numerals = "0.1.4"
 paste = "1.0.14"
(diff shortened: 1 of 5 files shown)


reknih:
typst/citationberg#142 is merged, so you can now add a [patch](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section) to the `Cargo.toml` so we can merge this with passing CI.

reknih:
Thank you!

MangoIV:
is this in any release typst version? I'm running into this problem but I cannot get this to actually use a single page...

reknih:
This will land in Typst 0.12 for which we expect a preview version this week. I will review this PR today.

PR #155, opened 2024-04-23
Parse page ranges following citeproc

DerDrodt:
This PR depends on typst/citationberg#7.

Allows for `page-range` field to have multiple ranges, as in `1-4 & 7-12, 13, 14\-19`.

Each of these is split into parts:
- Commas as `,`
- Ampersands like `&`
- Single pages like `13`
- Escaped ranges like `14\-19`, where the hyphen will not be replaced by the given separator
- Ranges like `1-4`

Ranges are given to citationberg to format depending on the format rule per the CSL spec.


reknih:
I have fixed the compile error and removed the provisional error message. I noticed that there were a few test regressions:

<details>
❌ Test label_EmptyLabelVanish should pass but failed<br>
❌ Test locale_PageRangeDelimiterTermDefined should pass but failed<br>
❌ Test locale_PageRangeDelimiterTermFrenchUndef should pass but failed<br>
❌ Test number_PageFirst should pass but failed<br>
❌ Test number_StrangeError should pass but failed<br>
❌ Test page_Chicago should pass but failed<br>
❌ Test page_Chicago16 should pass but failed<br>
❌ Test page_NumberPageFirst should pass but failed<br>
❌ Test page_PluralDetectWithEndash should pass but failed<br>
❌ Test page_WithLocaleAndWeirdDelimiter should pass but failed<br>
❌ Test virtual_PageFirst should pass but failed
<summary>citeproc regressions</summary>
</details>

Chicago, for example, seems to be reversed in terms of what numbers it omits. It would be great if you could check what's going wrong here and if there's a common cause!

Drodt:
There was a minor bug in the page range formats for citationberg that I fixed.

Besides that I put `page` into its own variable kind, as it behaves quite different from other variables. With these changes, all tests pass and I even managed to get one more to pass.

Look into the design changes, especially for citationberg, and let me know if you want something changed.

reknih:
I merged https://github.com/typst/citationberg/pull/14 with two comments added, thank you!

Drodt:
From my POV this is good to merge. There might be further work down the line but this is worth an update imo

src/csl/rendering/mod.rs
```diff
@@ -393,7 +392,11 @@ fn label_pluralization(
         LabelPluralize::Contextual => match variable {
             NumberVariableResult::Regular(MaybeTyped::String(_)) => false,
             NumberVariableResult::Regular(MaybeTyped::Typed(n)) => {
-                n.is_plural(label.variable.is_number_of_variable())
+                if let NumberOrPageVariable::Number(v) = label.variable {
+                    n.is_plural(v.is_number_of_variable())
+                } else {
+                    panic!("Incompatiable variable types")
+                }
```
reknih, line 399: It would be nice to enforce this on a type system level but it's okay for now

reknih (commented):
I have made a few changes to the comparison functions and the `resolve_page_variable` function. Thank you for your work!

PR #174, opened 2024-05-26
Fix not using short titles

pjtsearch:
This PR is to address issue #173, which was that hayagriva did not use the BibTex `shorttitle` field for short titles in CSL. In order to fix this issue, I made two changes:

1. In [interop.rs](https://github.com/typst/hayagriva/compare/main...pjtsearch:fix-short-titles?expand=1#diff-f2d7ad31a3f6f942aa18f5b56f944448cdf5f10c83d0ea16e39caf36157247c7), I added functionality for detecting if the BibTex `shorttitle` exists and adding it to the `FormatString` if it does.
2. In [taxonomy.rs](https://github.com/typst/hayagriva/compare/main...pjtsearch:fix-short-titles?expand=1#diff-702ef24022db1313b47f2d64bc684ce2aee610ab03badbf79cf6cb50eb692559), I fixed the code for `StandardVariable::Title` and `StandardVariable::TitleShort` being switched around. Previously, hayagriva would always select the short title for `StandardVariable::Title`, but would choose between the short title and long title for `StandardVariable::TitleShort`, which is obviously incorrect behavior. It appears that the behavior for `StandardVariable::TitleShort` was mistakenly placed in `StandardVariable::Title`. I fixed this issue by simply switching the code around.

### DIFF
--- a/src/csl/taxonomy.rs
+++ b/src/csl/taxonomy.rs
@@ -329,13 +329,13 @@ impl EntryLike for Entry {
                 .map(|f| f.select(form))
                 .map(Cow::Borrowed),
             StandardVariable::Status => None,
-            StandardVariable::Title => entry
+            StandardVariable::Title => {
+                entry.title().map(|f| f.select(form)).map(Cow::Borrowed)
+            }
+            StandardVariable::TitleShort => entry
                 .title()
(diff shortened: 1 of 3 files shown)


reknih:
Looks good, thank you!

PR #168, opened 2024-05-16
Fix sorting citation by citation number

lluchs:
Citations with citation numbers currently aren't sorted properly, for example output from the test in the second commit:

```
Expected:
[1]–[4]
[1]–[4]

Got:
[1]–[4]
[2], [1], [4], [3]
```

This PR defers sorting the citations until `finish` is called and the citation numbers are determined.

Fixes #115

PR #176, opened 2024-05-28
Fix formatting of collapsed numeric citations

### LINKED ISSUE #154: Incorrect formatting of collapsed in-text citations when using `collapse` and `after-collapse-delimiter` attribute in CSL
I have encountered an issue with the collapsing behavior of consecutive citation numbers in in-text citations when using the `bibliography` function in Typst. Using the citation style [`"gb-7714-2015-numeric"`](https://www.zotero.org/styles/china-national-standard-gb-t-7714-2015-numeric?source=1), the relevant CSL code for citation rendering is as follows:

```xml
<citation collapse="citation-number" after-collapse-delimiter=",">
  <layout vertical-align="sup" delimiter="," prefix="[" suffix="]">
    <text macro="citation-layout"/>
  </layout>
</citation>
```

According to this style, in-text citations should be rendered as superscript numbers enclosed in square brackets, e.g., `[1]`. If multiple references are cited at the same location, their numbers should be separated by commas, e.g., `[1,3,5]`. If the cited reference numbers are consecutive, they should be collapsed, e.g., `[1-3,5]`. Zotero also renders in-text citations in this manner.

However, in Typst, the collapsing behavior does not work as expected. When citing multiple consecutive references at the same location, Typst displays them as `[1,3]` instead of the expected `[1-3]`. Similarly, when citing references 1, 2, 3, and 5 at the same location, Typst renders them as `[1,3,5]` instead of the expected `[1-3,5]`. 

```typst
1 - @TestBook_1
2 - @TestBook_2
3 - @TestBook_3
4 - @TestBook_4
5 - @TestBook_5

Test 1: Expected #super[[1-3,5]] Actual @TestBook_1 @TestBook_2 @TestBook_3 @TestBook_5

Test 2: Expected #super[[1,3,5]] Actual @TestBook_1 @TestBook_3 @TestBook_5

Test 3: Expected #super[[1-3]] Actual @TestBook_1 @TestBook_2 @TestBook_3

#bibliography("ref.bib", style: "gb-7714-2015-numeric")
```
image

Through testing, I found that in Typst/hayagriva the `after-collapse-delimiter` attribute in the CSL file controls the delimiter used for collapsing consecutive numbers. Changing `after-collapse-delimiter=","` to `after-collapse-delimiter="-"` results in the expected behavior, displaying `[1-3,5]`.

However, according to the [CSL 1.0.2 documentation - Cite Collapsing](https://docs.citationstyles.org/en/stable/specification.html#cite-collapsing), the `after-collapse-delimiter` attribute is described as:

> Specifies the cite delimiter to be used _**after**_ a collapsed cite group.

Based on the example provided in the documentation, I think this attribute should control the comma (`,`) in `[1-3,5]`, not the hyphen (`-`). I have tested other numeric citation styles with the `collapse` attribute and observed similar issues.

I kindly request guidance on whether this behavior is intended or if there is a potential issue with the collapsing implementation in Typst/hayagriva.

bluebear94:
Fixes #154.


src/csl/archive.rs
```diff
@@ -217,7 +217,7 @@ pub enum ArchivedStyle {
     ///
     /// Authors: Patrick O'Brien.
     MaryAnnLiebertVancouver,
-    /// Modern Humanities Research Association 3rd edition (note with bibliography).
+    /// Modern Humanities Research Association 4th edition (note with bibliography).
```
PgBiel, line 220: Hi, can you please remove the following changes from the PR's diff:

1. All changes under `archive/styles`
2. Changes to `archive.rs`

They are currently automatically changed by hayagriva's tests, which is annoying (we will improve this), but for now we should just remove the diff from the PR until this is improved :)

src/csl/mod.rs
```diff
@@ -853,25 +853,39 @@ fn collapse_items<'a, T: EntryLike>(cite: &mut SpeculativeCiteRender<'a, '_, T>)
     match style.citation.collapse {
         Some(Collapse::CitationNumber) => {
             // Option with the start and end of the range.
+            let after_collapse_delim =
```
PgBiel, line 856: Might be worth moving this line upwards, as it is shared by the two non-empty match branches.

src/csl/mod.rs
```diff
@@ -2852,4 +2866,72 @@ mod tests {
             // }
         }
     }
+
+    #[test]
+    fn test_citation_collapsing() {
```
PgBiel, line 2871: Thanks to PR #168, we now have local citeproc-style unit tests at `tests/local`. Do you think you could create a test in `tests/local` instead? That should allow us to remove both this function and the `styles/numeric.csl` file (see the existing test there for a sample; see https://github.com/citation-style-language/test-suite for more information on the test format).

src/csl/mod.rs
```diff
@@ -853,25 +853,39 @@ fn collapse_items<'a, T: EntryLike>(cite: &mut SpeculativeCiteRender<'a, '_, T>)
     match style.citation.collapse {
         Some(Collapse::CitationNumber) => {
             // Option with the start and end of the range.
+            let after_collapse_delim =
+                after_collapse_delim.or(style.citation.layout.delimiter.as_deref());
             let mut range_start: Option<(usize, usize)> = None;
+            let mut ended_range = false;
```
PgBiel, line 859: I feel like this name needs the word `collapsed` somewhere. Something like `previous_range_collapsed` would be fairly descriptive, but maybe `just_collapsed` is better and less verbose.

PgBiel (changes requested):
Thank you, this is looking great! I have a few comments.

PgBiel:
Thank you! :blue_heart:

PR #205, opened 2024-08-27
Re-export `BibliographyItem`

### LINKED ISSUE #157: `BibliographyItem` is not re-exported
See here: https://docs.rs/hayagriva/latest/hayagriva/struct.RenderedBibliography.html

xkevio:
Fixes #157.

### DIFF
--- a/src/csl/mod.rs
+++ b/src/csl/mod.rs
@@ -1005,10 +1005,15 @@ pub struct RenderedBibliography {
     pub items: Vec<BibliographyItem>,
 }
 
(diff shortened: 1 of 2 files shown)


xkevio:
Oh yeah, adding docs would make sense whoops. Should I do that, or is that something you would do @PgBiel?

PgBiel:
> Oh yeah, adding docs would make sense whoops. Should I do that, or is that something you would do @PgBiel?

This one's on us, sorry about that.

Thank you for the PR! :blue_heart:

PR #210, opened 2024-09-21
docs: change example url access date

Tokosan:
Change the example date on `docs/file-format.md` from `2020-11-12` to `2020-11-30` to avoid confusion between day and month. Different countries have different date formats, so having both numbers beneath 12 might be confusing.


PgBiel (approved):
Thank you

PR #218, opened 2024-10-01
Add note field to interop

Drodt:
The `note` field of biblatex was not added to the note field of hayagriva during conversion. This PR fixes this.

### DIFF
--- a/src/interop.rs
+++ b/src/interop.rs
@@ -529,6 +529,12 @@ impl TryFrom<&tex::Entry> for Entry {
             }
         }
 
+        if let Some(note) = map_res(entry.note())?.map(|d| d.format_verbatim()) {
+            if item.note.is_none() {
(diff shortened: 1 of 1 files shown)


EpicEricEE:
Should this maybe rather replace the `annotation` in the block above? The annotation is later used for the `annote` field after all so it doesn't really make sense to have it twice.

Drodt:
> Should this maybe rather replace the `annotation` in the block above? The annotation is later used for the `annote` field after all so it doesn't really make sense to have it twice.

Looks like you're right. But I would like confirmation by the main developers. I don't want to break things.

Drodt:
> By the way @Drodt, do you think you can update #105? I like the change and we will release a new Hayagriva version tomorrow.

It can be merged without errors now.

reknih:
See #220

PR #228, opened 2024-10-08
Add support for subsequent author substitution

### LINKED ISSUE #219: subsequent-author-substitute is not respected
Currently, hayagriva ignores the `subsequent-author-substitute` option on the bibliography. It should follow the CSL specification: https://docs.citationstyles.org/en/stable/specification.html?highlight=label#reference-grouping

Drodt:
Fixes #219.

Also fixes the error that affixes for bibliography layout were ignored. And updates archives.

Adds 4 tests to `citeproc-pass.txt`.

Still a draft until:
- [x] Code is reworked to be a bit nicer
- [x] Tests for bibliography are supported (See #229).
- All relevant tests pass:
    - [x] name_SubsequentAuthorSubstituteSingleField
    - [x] name_SubstitutePartialEach
    - [x] name_SubsequentAuthorSubstituteMultipleNames


Drodt:
If possible, this should be merged for typst v0.12, in my opinion.

archive/locales/de-AT.cbor
PgBiel, line 1: Please undo changes to archive files, they should be updated separately. (You can rebase on main to get the latest updates.) Thank you!

tests/citeproc.rs
```diff
@@ -303,6 +310,39 @@ impl<'s> TestCaseBuilder<'s> {
     }
 }
 
+fn extract_from_html(html: &str) -> String {
```
PgBiel, line 313: Could you implement bibliography test support on a separate PR? As is, it's harder to tell at a glance which changes were due to subsequent author substitution support and which changes were due to newly-supported bibliography tests.

tests/f219.rs
```diff
@@ -0,0 +1,85 @@
+use hayagriva::citationberg::IndependentStyle;
```
PgBiel, line 1: I do wonder if this could become a `local` citeproc test with the new bibliography support.

Though it'd generally be nice to have better infrastructure for non-citeproc (hayagriva YAML) tests. I wonder if we could have a nice little macro which does all of this in a less repetitive way, so we can have multiple tests in a single file.

tests/f219.rs
```diff
@@ -0,0 +1,85 @@
...
+    ));
+
+    let result = driver.finish(BibliographyRequest {
+        style: &style,
+        locale: None,
+        locale_files: &locs,
+    });
+
+    for i in result.bibliography.unwrap().items {
+        eprintln!("{}", i.content);
```
PgBiel, line 83: (I also believe this should be an assertion.)

PgBiel (commented):
Thanks, here are some initial thoughts.

reknih:
I have merged main in here, removed the HTML testing in favor of #273, and  got rid of the archive updates.

Two tests are failing now:
- `bugreports_parenthesis`, first passed in your PR before my merge
- `substitute_SuppressOrdinaryVariable`, first passed in https://github.com/typst/hayagriva/commit/1058be7dab3d34bb317e26360d79b2830cb391ed on main

Drodt:
> I have merged main in here, removed the HTML testing in favor of #273, and got rid of the archive updates.

Thanks

> Two tests are failing now:
> 
>     * `bugreports_parenthesis`, first passed in your PR before my merge

The issue here is that citeproc removes collapses two spaces following each other, while hayagriva does not. Does hayagriva do any of the other such simplifcations that citeproc does, e.g., transforming `!.` to `!`?

>     * `substitute_SuppressOrdinaryVariable`, first passed in [1058be7](https://github.com/typst/hayagriva/commit/1058be7dab3d34bb317e26360d79b2830cb391ed) on main

Here the problem is that the period following the title is not moved into the quotation marks. I find that odd, as I did not touch that logic.

Drodt:
I removed `bugreports_parseName` in favor of opening #304

Drodt:
This PR is also ready, once you have the time.

PgBiel:
Big sorry for the delay by the way, I've been incrementally reviewing this PR for several weeks now, but there's quite a bit going on and I wanted to think it through :sweat_smile: 
I'll try to conclude the first round soon.

PR #231, opened 2024-10-09
Fix #227

### LINKED ISSUE #227: Hayagriva 0.7 breaks due to MaybeTyped
The latest main release of typst, featuring Hayagriva 0.7, encounters errors whenever it attempts to load unused YAML entries. This issue did not occur in Hayagriva 0.6, suggesting a strong likelihood that the bug is related to the recent update in Hayagriva.

image

reknih:
With the new page range type, `Deserialize` only returned `Ok` if the value was a string. This PR adds conversions for i32, u32, and u64 values which YAML can easily yield for single pages

Fixes #227

### DIFF
--- a/src/types/page.rs
+++ b/src/types/page.rs
@@ -2,8 +2,8 @@ use std::{cmp::Ordering, fmt::Display, num::NonZeroUsize, str::FromStr};
 
 use crate::{MaybeTyped, Numeric, NumericError};
 
-use super::{deserialize_from_str, serialize_display};
-use serde::{de, Deserialize, Serialize};
+use super::{derive_or_from_str, serialize_display};
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
 impl MaybeTyped<PageRanges> {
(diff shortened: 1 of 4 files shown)


src/types/page.rs
```diff
@@ -23,6 +28,34 @@ pub struct PageRanges {
     pub ranges: Vec<PageRangesPart>,
 }
 
+derive_or_from_str!(@deser_impl PageRanges where "pages, page ranges, ampesands, and commas",
```
EpicEricEE, line 31:
```suggestion
derive_or_from_str!(@deser_impl PageRanges where "pages, page ranges, ampersands, and commas",
```

src/types/page.rs
```diff
@@ -116,6 +173,44 @@ pub enum PageRangesPart {
...
+
+        #[derive(Deserialize)]
+        #[serde(rename_all = "kebab-case")]
+        enum Inner {
+            Ampersand,
+            Comma,
+            EscapedRange(Numeric, Numeric),
+            SinglePage(Numeric),
+            Range(Numeric, Numeric),
+        }
```
PgBiel, line 189: Out of curiosity, is this deserializing something like

```jsonc
{
  // PageRanges
  "ranges": [
     // PageRangesPart::Ampersand
     "ampersand",
     // PageRangesPart::Comma
     "comma",
     // PageRangesPart::EscapedRange
     { "escaped-range": [12, 34] }
  ]
}
```
?
Is this used somewhere?
(I'm aware the FromStr impl covers the usual cases)



Looking further around, this seems to be just for consistency with the rest of types, so that seems acceptable.

src/types/page.rs
```diff
@@ -169,9 +264,36 @@ impl PageRangesPart {
...
+        Ok(Self::SinglePage(value.into()))
+    }
+}
+
+impl TryFrom<u64> for PageRangesPart {
+    type Error = TryFromIntError;
+
+    fn try_from(value: u64) -> Result<Self, Self::Error> {
+        let value: i32 = value.try_into()?;
+        Ok(Self::SinglePage(value.into()))
```
PgBiel, line 296: Removing `From<u64>` (which, at a glance, appears to be the correct decision) caused a compilation error in `csl-json` when resolving page variables, as `PageRanges::From(u64)` no longer works:

https://github.com/typst/hayagriva/blob/a1a27dc6ac58ddde3b4deed3ab441619d9e2b25a/src/csl/taxonomy.rs#L807-L815

Is doing a fallible conversion and returning `None` on error an appropriate choice here? Not sure if that's how `Option` is supposed to be interpreted in this particular internal API. We could also just saturate specifically here and call it a day, though I'd be inclined to indicate failure somehow.

PgBiel (commented):
Seems OK in general. I'll leave a few comments.

PR #244, opened 2024-10-27
Apostrophe should not always cause step to return CharClass::InWord

### LINKED ISSUE #196: @Citation makes word following `'` lowercase in escaped Hayagriva entry
### Description

Hayagriva entry:  

```
ukraine: 
  type: video
  title: "'Russia Is Not Going To Lose The War': JD Vance Blasts Senate Voting To Send Funding To Ukraine"
  date: 2024-02-14
  url: 
    value: https://www.youtube.com/watch?v=W53wILyEyEc
    date: 2024-08-02 
```

document.typ: 

```
Blah blah blah @ukraine 
```

Expected result: The title should appear as 'Russia Is Not Going To Lose The War': JD Vance Blasts Senate Voting To Send Funding To Ukraine

Result (regardless of citation style): The title appears as 'russia Is Not Going To Lose The War': JD Vance Blasts Senate Voting To Send Funding To Ukraine

Note: With some experimentation, I have learned that this happens regardless of where the `'` is. Any character following `'` is lowercase. 

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

sandbergja:
Similar to `"`, `”`, `«`, and other such characters, an apostrophe can mark a quotation, in which case the character after it may need to be upper case. Before this commit, Hayagriva always considered the character after an apostrophe to be within a word, and therefore lowercased it.

Closes #196

### DIFF
--- a/src/lang/mod.rs
+++ b/src/lang/mod.rs
@@ -164,7 +164,7 @@ impl CharClass {
             }
             '(' | '[' | '{' => Self::NewSubclause,
             ')' | ']' | '}' => self,
-            '"' | '”' | '“' | '»' | '›' | '«' | '‹' | '‘' => self,
+            '"' | '”' | '“' | '»' | '›' | '«' | '‹' | '‘' | '\'' => self,
             ',' | ';' => {
                 if self.is_in_word() {
                     Self::MaybeSubclause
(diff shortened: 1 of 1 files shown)


sandbergja:
Rebased this so that hopefully CI passes. :-)

PgBiel:
This seems legitimate to me. Thanks!

PR #245, opened 2024-10-27
Update archived locale files

sandbergja:
The tests were failing for me locally, but running

```
HAYAGRIVA_ARCHIVER_UPDATE=1 cargo test
```

worked.  Here are the file updates I received from that, in case they are useful for others.


PgBiel:
Don't worry, we have updated them and will be updating from time to time. We plan on improving this aspect of our CI in the future so it's more automatic, but for now you'll just have to let us know it is outdated. Thank you! :)

PR #248, opened 2024-11-11
Fix --format CLI flag

m-haug:
Previously, the CLI used clap's PossibleValueParser, which only supports parsing strings. However, hayagriva was expecting to parse Format directly. This caused a panic in clap, when trying to convert the String to Format.

With this patch, Format implements ValueEnum, so we can use clap::value_parser to create an EnumValueParser to parse the Format.


PgBiel:
Thank you!

PR #250, opened 2024-11-20
Fix unnecessary disambiguation around ibids

### LINKED ISSUE #129: Possible issues with disambiguation?
Does somebody know how I can understand the issue if disambiguation is applied incorrectly?

Given a CSL with such an entry:
`<citation et-al-min="4" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" givenname-disambiguation-rule="primary-name" collapse="year">`

I'm seeing an entry such as 2016b, yet it's the only entry for 2016, the only entry for that author and there's no 2016a.

Any idea on how I can understand the root cause?

Example see:
image

MaxGietl:
Fixes https://github.com/typst/hayagriva/issues/129.

Disambiguation currently compares the normal render of a citation. This means that if two different citations both render to "ibid", they are disambiguated without actually being ambiguous. This PR solves that by creating a second render that disables ibids and using that for comparison.

Example:
```typst
First mention of Smith @smith[1].
Second mention of Smith @smith[2].

First mention of Mayer @mayer[1].
Second mention of Mayer @mayer[2].

Hayagriva disambiguates the citations, even though they aren't ambigous. 

#bibliography("bibliography.yaml", style: "deutsche-sprache")
```
bibliography.yaml
```yaml
smith:
  type: article
  title: First article of Smith
  author: Smith
  date: 1990

mayer:
  type: article
  title: Article of Mayer
  author: Mayer
  date: 2010
```

Faulty render without this PR
Screenshot 2024-11-20 154755

Correct render with this PR
image


MaxGietl:
@tsteckenborn Could you confirm this also fixes your issue or could you provide the files that caused this for you?

PgBiel:
> These tests also fail on main currently, so I think this has nothing to do with my changes?

Are you sure? I'd expect archive tests to fail as archive tests just compare the existing archive files with upstream CSL styles, but no citeproc tests are failing on main (at least, the CI check is green). Your PR might have triggered some unintended side-effects (would need further investigation).

PR #254, opened 2024-11-28
Fix Markdown warnings in README: multiple top-level headers and code blocks without surrounded blank lines

bact:
- There should be only one H1 heading in a Markdown file. Fix H1 "Usage" to H2.
  - See: https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md025---multiple-top-level-headers-in-the-same-document 
- There should be one blank line before the code block. Add blank lines.
  - See: https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md031---fenced-code-blocks-should-be-surrounded-by-blank-lines

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Hayagriva supports all styles provided in the
 [official Citation Style Language repository](https://github.com/citation-style-language/styles),
 currently over 2,600.
 
-# Usage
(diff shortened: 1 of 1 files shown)


PgBiel:
Thanks!

PR #150, opened 2024-04-05
fix: handle initials correctly

razvanazamfirei:
This PR fixes the handling of initials when no delimiter is provided. Previously, citation styles that initialized without a delimiter (e.g., Vancouver) were formatted incorrectly. 

For example:
> **Doe J J.** World's Best Paper. Journal. 2025;99(9): 1–99.

instead of 
> **Doe JJ.** World's Best Paper. Journal. 2025;99(9): 1–99.

To fix this, we no longer add a space if no delimiter is passed.

Fixes #137.

PR #257, opened 2024-12-08
Fix handling of trailing whitespaces in initials' delimiters

### LINKED ISSUE #137: Bibliography incorrectly adds a space between author initials
### Description

Just as an example I used the style "vancouver-superscript". It works with most styles though. 
 
Bildschirmfoto 2024-02-17 um 22 18 02

According to the official style, there shouldn't be a space between the "R" and the "G", so it should be "Astley RG" and not "Astley R G".

### Reproduction URL

_No response_

### Operating system

Web app

### Typst version

PgBiel:
Fixes #137

Some styles, such as `vancouver`, require there to be nothing between initials, so `Astley, Rick Gavin` would be formatted as `Astley, RG` instead of `Astley, R. G.` (more usual) or `Astley, R G` for example. However, even when setting `initializer-with=""` (empty delimiter), Hayagriva was still forcing whitespaces between initials.

It was thought that #150 would have fixed it by only adding whitespaces if a delimiter was indeed specified, but it actually didn't change anything as **there is always a delimiter** - however, it is sometimes **an empty string**. While adding an explicit check for empty strings would have fixed this specific problem, it turns out Hayagriva's behavior of forcing a whitespace to always be added after the delimiter, replacing any of its trailing whitespaces if there are any (otherwise, one is added anyways), is not consistent with other CSL processors. Indeed, several citeproc tests demonstrated that the expected way to add whitespaces between initials is through `initializer-with=". "` instead of `initializer-with="."`. So, I removed Hayagriva's automatic whitespace behavior, **except** when we're replacing hyphens (e.g. for `initializer-with=". "`, `Aa-Bb` should become `A.-B.` and not `A. -B.`).

It is, however, unfortunate that this seems to be somewhat inconsistent with a specific example given by the relevant section of the CSL 1.0.2 spec:

```md
`initialize-with`
    When set, given names are converted to initials. The attribute value is added after each initial (“.” results in “J. J. Doe”).
```

...but most official styles I've seen appear to assume that whitespaces are preserved, using `". "` instead of `"."` in that case. So this seems to be the way to go.

I've also added some tests. Two more citeproc tests are now passing as well.

I'll be keeping this as a draft as I still need to test the case with `initialize="false", initialize-with="."`, but in principle the code seems to preserve whitespace here as well, however whitespaces from the name itself are also kept, which is why `initialize-with="."` is used instead of `initialize-with=". "` by official styles in this case.

### DIFF
--- a/src/types/persons.rs
+++ b/src/types/persons.rs
@@ -234,8 +234,14 @@ impl Person {
     /// Formats the given name into initials.
     ///
     /// For example, `"Judith Beatrice"` would yield `"J. B."` if the
-    /// `delimiter` argument is set to `Some(".")`, `"Klaus-Peter"` would become
-    /// `"K-P"` without a delimiter.
+    /// `delimiter` argument is set to `Some(". ")`, `"Klaus-Peter"` would
+    /// become `"K-P"` without a delimiter (or with an empty delimiter).
+    ///
+    /// Whitespace at the end of the delimiter is kept when parts are separated
+    /// by whitespace (as in the first example), but not when parts of a
+    /// compound given name are separated by hyphens (for example,
+    /// `"Klaus-Peter"` produces `"K.-P."` instead of `"K. -P."` for a
+    /// delimiter of `Some(". ")`).
     pub fn initials(
         &self,
         buf: &mut impl std::fmt::Write,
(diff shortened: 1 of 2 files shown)


reknih:
If the other implementations and styles assume/exhibit/rely on this behaviour, we should also add it, so I'm good with it from that perspective.

PgBiel:
Taking a look here, it seems that automatic whitespaces should also be removed when `initialize="false"` but `initialize-with` is set, according to citeproc tests, whereas we are not doing so. But I'll leave that to a separate PR.

Issue #260, opened 2024-12-13
Cannot specify PMID/PMCID through BibLaTeX input


RyanVidegar-Laird:
Hi,
Some scientific grants require PMCID or PMID fields in citations. They are defined as standard variables in the [CSL spec](https://docs.citationstyles.org/en/stable/specification.html?#appendix-iv-variables). Zotero provides a nice CSL for [NLM grant proposals](https://www.zotero.org/styles/national-library-of-medicine-grant-proposals) that works fine creating a bib with its GUI.

However, PMCID/PMIDs are not shown when using the above CSL file in Typst. I've tried in the online app (example: https://typst.app/project/rFgaiSlKEcITlBZVkRVwkD), and locally using Typst version 0.12.0.

Seems related to #91 and #162. My naive take... the  respective variables are defined in [src/csl/taxonomy.rs](https://github.com/typst/hayagriva/blob/ff7bb08097797e4315a662b3f9e68d7c182264ce/src/csl/taxonomy.rs#L328), but *not* in [src/interop.rs](https://github.com/typst/hayagriva/blob/ff7bb08097797e4315a662b3f9e68d7c182264ce/src/interop.rs).

Thanks for all the dev effort! Please let me know if you need any more info.

Drodt:
The PMID/PMCID fields are not supported in biblatex (or bibtex). See (§3.14.7 of the [biblatex documentation](https://ftp.mpi-inf.mpg.de/pub/tex/mirror/ftp.dante.de/pub/tex/macros/latex/contrib/biblatex/doc/biblatex.pdf) for how you should encode this in biblatex.

Your example works fine if you write a hayagriva file directly.

RyanVidegar-Laird:
Thanks for the quick fix and response! I can confirm I'm able to include PMID and PMCIDs when using a haragriva file. To note for you and/or future users:
- When exporting a bibliography in Zotero, the default BibLaTex and BibTeX options _do not_ place PMIDs in the eprint field. I had to use the Better BibTex extension's 'Better BibLaTex' option. This was part of my issue.
- NIH training grants ([see page 43 in the instructions](https://grants.nih.gov/grants/how-to-apply-application-guide/forms-h/training-forms-h.pdf)) prefer PMCIDs over PMIDs when available, but articles with PMCIDs will always also have a PMID.
    - When using a hayagriva file this logic from the above CSL file works if both are included in the `serial-number` field. 
[...]

PR #261, opened 2024-12-15
Add translation for pubmed eprint

Drodt:
As in §3.14.7 of the [biblatex documentation](https://ftp.mpi-inf.mpg.de/pub/tex/mirror/ftp.dante.de/pub/tex/macros/latex/contrib/biblatex/doc/biblatex.pdf).

Addresses #260.

### DIFF
--- a/src/interop.rs
+++ b/src/interop.rs
@@ -583,3 +583,25 @@ fn comma_list(items: &[Vec<Spanned<Chunk>>]) -> FormatString {
 
     FormatString { value, short: None }
 }
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn test_pmid_from_biblatex() {
+        let entries = crate::io::from_biblatex_str(r#"@article{test_article,
+	title = {Title},
+	volume = {3},
+	url = {https://example.org},
+	pages = {1--99},
+	journaltitle = {Testing Journal},
+	author = {Doe, Jane},
+	date = {2024-12},
+ eprint = {54678},
+ eprinttype = {pubmed},
+}"#).unwrap();
+        let entry = entries.get("test_article").unwrap();
+        assert_eq!(Some("54678"), entry.keyed_serial_number("pmid"));
+        assert_eq!(Some("54678"), entry.pmid());
+    }
+}
+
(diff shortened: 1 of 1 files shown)


archive/styles/apa.cbor
PgBiel, line 1: Please undo this change (rebase on main). We're supposed to update this ourselves. :)

src/interop.rs
```diff
@@ -415,11 +415,12 @@ impl TryFrom<&tex::Entry> for Entry {
...
-            if map_res(entry.eprint_type().map(|c| c.format_verbatim().to_lowercase()))?
-                .as_deref()
-                == Some("arxiv")
-            {
+            let eprint_type =
+                map_res(entry.eprint_type().map(|c| c.format_verbatim().to_lowercase()))?;
+            if eprint_type == Some("arxiv".to_string()) {
                 item.set_arxiv(eprint);
+            } else if eprint_type == Some("pubmed".to_string()) {
+                item.set_pmid(eprint.replace("http://www.ncbi.nlm.nih.gov/pubmed/", ""));
```
PgBiel, line 423: We could remove the `to_string` calls like so:

```suggestion
            let eprint_type =
                map_res(entry.eprint_type().map(|c| c.format_verbatim().to_lowercase()))?
                .as_deref();
            if eprint_type == Some("arxiv") {
                item.set_arxiv(eprint);
            } else if eprint_type == Some("pubmed") {
                item.set_pmid(eprint.replace("http://www.ncbi.nlm.nih.gov/pubmed/", ""));
```

PgBiel, line 423: I don't think this logic is correct. The reference link you mentioned in the issue (https://ftp.mpi-inf.mpg.de/pub/tex/mirror/ftp.dante.de/pub/tex/macros/latex/contrib/biblatex/doc/biblatex.pdf) shows that `eprint` should contain the identifier directly, not the URL. The URL example was actually using the `url = ...` field and was supposed to exhibit the "bad" case, where the "good" case is when you write `eprint = {IDENTIFIER_HERE}; eprinttype = {pubmed}`. So we don't need to (and shouldn't) replace the URL here, as it shouldn't be there.

PgBiel (changes requested):
Thanks!

PR #268, opened 2025-01-22
Fix typo

YDX-2147483647:
Only a panic is visible to end users. Just fix them all together.


YDX-2147483647:
Test fail because of outdated archives.

> locale 'de-AT.cbor' is outdated, run archiver tests with env var 'HAYAGRIVA_ARCHIVER_UPDATE=1' to update

PgBiel:
Thank you!

PR #269, opened 2025-01-22
Fix delimiting within `<choose>` and `<layout>` for CSL

### LINKED ISSUE #109: Issue with CSL Rendering (OSCOLA no ibid Style)
### Description

OSCOLA style should have a comma between the edition and the publisher.

A valid rendering from https://editor.citationstyles.org/visualEditor/ (using [this csl json to get the citation example](https://github.com/typst/typst/files/13683538/csl.json)):
image

Compared to the rendering in typst.app:
image

Notice the missing comma.

The CSL style is https://github.com/citation-style-language/styles/blob/master/oscola-no-ibid.csl

### Reproduction URL

https://typst.app/project/rY0Nmyq6PphmJZy877_lLT

### Operating system

Web app, Windows

### Typst version

### LINKED ISSUE #180: The `delimiter` attribute of `<group>` is not applied to items inside `<choose>`
Consider the following CSL snippet:

```xml
<group delimiter=". ">
  <text value="v0"/> 
  <text value="v1"/>
  <choose>
    <if type="book">
      <text value="v2"/>
      <text value="v3"/>
    </if>
  </choose>
</group>
```

<details>
<summary> The complete test CSL file </summary>

```xml
<?xml version="1.0" encoding="utf-8"?>
<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="in-text" name-as-sort-order="all" sort-separator=" " demote-non-dropping-particle="never" initialize-with=" " initialize-with-hyphen="false" page-range-format="expanded" default-locale="zh-CN">
  <info>
    <title>test</title>
    <id>http://test</id>
    <category citation-format="numeric"/>
    <category field="generic-base"/>
  </info>
  <macro name="entry-layout">
    <group delimiter=". ">
      <text value="v0"/> 
      <text value="v1"/>
      <choose>
        <if type="book">
          <text value="v2"/>
          <text value="v3"/>
        </if>
      </choose>
    </group>
  </macro>
  <citation collapse="citation-number" after-collapse-delimiter=",">
    <sort>
      <key variable="citation-number"/>
    </sort>
    <layout vertical-align="sup" delimiter="," prefix="[" suffix="]">
      <text variable="citation-number"/>
    </layout>
  </citation>
  <bibliography>
    <layout>
      <text variable="citation-number" prefix="[" suffix="]"/>
      <text macro="entry-layout"/>
    </layout>
  </bibliography>
</style>
```

</details>

Expected reference output: `v0. v1. v2. v3`  
Actual reference output in Typst: `v0. v1. v2v3`

image

I noticed that when `<group>` contains `<choose>`, the group delimiter is not applied to elements inside `<choose>`. Instead, `<choose>` is treated as a group without a delimiter. 

This behavior is inconsistent with reference management software like Zotero. In the CSL 1.0.2 documentation, it is stated:
> Delimiters from the nearest delimiting element are applied within the output of `cs:choose` (i.e., the output of the matching `cs:if`, `cs:else-if`, or `cs:else`; see delimiter).
>
> src: https://docs.citationstyles.org/en/stable/specification.html#choose

---

P.S. Currently, this issue can be mitigated by nesting `<group delimiter=". ">` inside `<choose>`. However, this means that additional modifications to the CSL are required for it to work with Typst.

<details>
<summary> Example </summary>

```xml
<group delimiter=". ">
  <text value="v0"/> 
  <text value="v1"/>
  <choose>
    <if type="book">
      <group delimiter=". "> <!-- Add a nested group -->
        <text value="v2"/>
        <text value="v3"/>
      </group>  <!-- Add a nested group -->
    </if>
  </choose>
</group>
```

</details>

YDX-2147483647:
- Fix delimiting within `<choose>` and `<layout>` for CSL

	Fixes #180

    Fixes #109

	- Before: `Frank Mittelbach and others, The Latex Companion (2nd ednAddison-Wesley Professional 2004).`
	- After: `Frank Mittelbach and others, The Latex Companion (2nd edn, Addison-Wesley Professional 2004).`
	- Diff: The comma after `2nd edn`.

- Add a minimal test case

	- Before: `Title_Edition_PublisherPlace_TitleEditionPublisherPlace`
	- After: `Title_Edition_Publisher_Place_TitleEditionPublisherPlace`
	- Diff: `_` between the first _Publisher Place_.

There is a design trade-off:
The delimiter stack uses `String` instead of `&str`.
Using `&str` could eliminate many cloning, but at the cost of introducing complex lifetimes.
Considering that delimiters are typically short, I choose `String`.

Besides, I bump some `*.cbor` to pass tests. If this is not desired, I can reset the commit.


PgBiel (approved):
Thank you!

YDX-2147483647:
You are welcome!

And I notice that you updated the doc comments for push/pop (7a341cc139947e470241b50b858ad021f7d3102e).

> (the delimiter) is inherited by [`citationberg::Choose`] for its _children_ as well.

The world might be a bit confusing. In XML, the term_child_ only includes the _direct_ ones. (The term for both direct and indirect ones is _descendant_.)
The spec uses _output_ instead:

> [delimiting the _output_ of the child elements](https://docs.citationstyles.org/en/v1.0.2/specification.html#delimiter)

> [applied within the _output_ of `cs:choose`](https://docs.citationstyles.org/en/v1.0.2/specification.html#choose)

PgBiel:
It is correct in the sense that it only passes the delimiter forward to children, which may decide to pass it further down the tree or not - usually not, as the point of delimiter is to delimit only the immediate children of a group or layout, unless the child of the `choose` is another `choose`. The usage of the word "output" seems to be there more to indicate that it won't apply to, say, the `<else>` branch if it wasn't matched, but otherwise it will apply to the final children.

So I guess it's more a matter of how you interpret it - it's true that the delimiter may affect descendants other than children with nested `<choose>`, but I also wasn't looking to summarize how the children handle delimiters.

PR #284, opened 2025-03-03
Use roman-numerals-rs

AA-Turner:
The [``numerals``](https://crates.io/crates/numerals) crate was last updated five years ago, this PR adopts a no-std / 0BSD crate that is also faster.

`roman-numerals-rs` implements [`TryFrom`](https://docs.rs/roman-numerals-rs/latest/roman_numerals_rs/struct.RomanNumeral.html#impl-TryFrom%3Ci32%3E-for-RomanNumeral), so the `match` construct can be simplified (at the cost of minor duplication).

A


AA-Turner:
None of the lint or test failures seem related to this PR:

```
error: rustc 1.80.0 is not supported by the following packages:
  litemap@0.7.5 requires rustc 1.81
  zerofrom@0.1.6 requires rustc 1.81
```

```
failures:

---- always_archive stdout ----

thread 'always_archive' panicked at tests/archiver.rs:117:73:
called `Result::unwrap()` on an `Err` value: Custom("unknown variant `page`, expected one of `chapter-number`, `citation-number`, `collection-number`, `edition`, `first-reference-note-number`, `issue`, `locator`, `number`, `number-of-pages`, `number-of-volumes`, `page-first`, `part-number`, `printing`, `printing-number`, `section`, `supplement-number`, `version`, `volume`")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    always_archive
```

PR #287, opened 2025-03-05
Propagate font-variant from name element to name parts

### LINKED ISSUE #175: `font-variant="small-caps"` not supported for `<name>`
Hi folks,

I want to make a minor adjustment to the IEEE CSL definition, basically just having all author names in small caps. However, if I use small-caps as the `@font-variant` of the `<name>` element itself, Typst does not render it correctly, it gets ignored.
When I move it up in the hierarchy to `<names>` it works, but will also show non-name words like "and" in small caps which I don't want.

Here is a screenshot showing that small-caps is ignored (see line 115):
image

sonovice:
When using font-variant="small-caps" on the <name> element in CSL styles, this formatting wasn't properly inherited by the individual name parts, causing the styling to be ignored. This fix ensures that font-variant styling is correctly propagated from the parent <name> element to child name parts (given name and family name) while maintaining the behavior where connector words like "and" remain in normal text.

This allows for proper small-caps styling of author names in bibliography styles like IEEE while keeping connector words in regular formatting.

image

Fixes #175.

### DIFF
--- a/src/csl/rendering/names.rs
+++ b/src/csl/rendering/names.rs
@@ -596,25 +596,11 @@ fn write_name<T: EntryLike>(
 
     // Combine the name formatting with specific part formatting
     let first_format = first_part
-        .map(|p| {
-            let mut fmt = p.formatting;
-            // If the name has font-variant set but the part doesn't, inherit it
-            if name_formatting.font_variant.is_some() && fmt.font_variant.is_none() {
-                fmt.font_variant = name_formatting.font_variant;
-            }
-            fmt
-        })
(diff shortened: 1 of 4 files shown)


sonovice:
(Checks seem to be unrelated to this PR.)

PgBiel:
> (Failed checks seem to be unrelated to this PR.)

Fixed. (Haven't taken a look at the PR itself yet.)

PgBiel:
From your solution, ended up finding a simpler one. Thank you!

PR #289, opened 2025-03-07
Bump citationberg to fix page variable in Number

PgBiel:
Necessary to update archives.

### DIFF
--- a/src/csl/rendering/mod.rs
+++ b/src/csl/rendering/mod.rs
@@ -19,7 +19,7 @@ use crate::lang::{Case, SentenceCase, TitleCase};
 use crate::types::{ChunkedString, Date, MaybeTyped, Numeric};
 use crate::PageRanges;
 
-use super::taxonomy::EntryLike;
+use super::taxonomy::{EntryLike, NumberOrPageVariableResult};
 use super::{write_year, Context, ElemMeta, IbidState, SpecialForm, UsageInfo};
 
 pub mod names;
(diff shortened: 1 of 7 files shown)


reknih (approved):
Looks good to me. The enum names are getting a bit unwieldy / long though.

src/csl/rendering/mod.rs
```diff
@@ -312,25 +335,36 @@ impl RenderCsl for citationberg::Number {
...
             Some(SpecialForm::OnlyFirstDate | SpecialForm::OnlyYearSuffix) => {
-                return self.variable == NumberVariable::Locator
+                return matches!(
+                    self.variable,
+                    NumberOrPageVariable::Number(NumberVariable::Locator)
+                )
             }
             Some(SpecialForm::VarOnly(_)) => return false,
             _ => {}
         }
```
PgBiel, line 365: I feel like something is off in this function. Seems like VarOnly forms will always return `false` here, even if the variables match.

My change at L353 is actually wrong since, in theory, a matching page variable should also pass. But since VarOnly always fails, it doesn't matter. However, since it appears that VarOnly is currently only used for author variables, this potential bug has no practical effect. Either way, do we know if this is intentional? @reknih

Could always fix it later, but just curious as i just noticed this. :)

PgBiel:
> Looks good to me. The enum names are getting a bit unwieldy / long though.

yeah... maybe we can come up with something shorter and more creative for "NumberOrPage", like "Numberlike" or similar.

src/csl/rendering/mod.rs
PgBiel: Since this effectively isn't a regression, I'll be merging this and we can fix in a separate PR if there is a bug here.

PR #296, opened 2025-03-17
biblatex `type` to `genre`, `number` to `serial` for non-article.

### LINKED ISSUE #295: Report entries with the bibtex 'number' field are translated to 'issue' variable
Bibtex/biblatex has a `number` field, which may or may not be numeric. Hayagriva translates both `number` and `issue` to `issue`.  CSL also has a `number` field, giving the example of a report number; while this is a number variable, the CSL spec appears to allow the variable to be non-numeric content.

In my case, I had been trying to handle this technical report:

```bibtex
@techreport{winfree_simulations_1998,
  title = {Simulations of Computing by Self-Assembly},
  author = {Winfree, Erik},
  year = {1998},
  number = {CSTR1998.22},
  institution = {Caltech}
}
```

Here, it seems like `number` should be translated to `serial-number` in Hayagriva's format, and in turn into `number` in CSL, rather than issue. Additionally, 'type' likely needs to be translated to 'genre'.  

As an example of this problem, the CSL for 'american-chemical-society' includes the following for a report:

```xml
            <choose>
              <if type="report">
                <group delimiter=" ">
                  <text variable="genre"/>
                  <text variable="number"/>
                </group>
              </if>
            </choose>
```

cgevans:
This is meant to close #295 :

The `type` field (not entry type) in biblatex is used to set `genre`, regardless of entry type.  I think this is appropriate for all uses of `type`; biblatex's documentation states `type` is "The type of a manual, patent, report, or thesis. This field may also be useful for the custom types listed in § 2.1.3.", while Hayagriva's file format states `genre` is "Type, class, or subtype of the item (e.g. "Doctoral dissertation" for a PhD thesis; "NIH Publication" for an NIH technical report). Do not use for topical descriptions or categories (e.g. "adventure" for an adventure movie).".

The `number` field is more of a problem.  Biblatex's documentation describes it as "The number of a journal or the volume/number of a book in a series."  However, it then goes on to say "With `@patent` entries, this is the number or record token of a patent or patent request", and, not documented in the description of `number`, is listed as an optional field in several entry types that seem more likely to be using it as a CSL number / Hayagriva serial-number than an issue number, namely report, manual, and dataset.  So I'm not quite sure what the best handling here is.   

I've done what seems to be to be the minimal change, which is to set serial-number for Repository, Reference, Patent, and Report, the Hayagriva types corresponding to the biblatex types that list number as an optional field where serial-number seems to make more sense.  But I wonder whether non-default styles also use it for other types, eg, Manuscript, Thesis, etc?

### DIFF
--- a/src/interop.rs
+++ b/src/interop.rs
@@ -393,6 +393,13 @@ impl TryFrom<&tex::Entry> for Entry {
             }
         }
 
+        // "number" is generally used in Biblatex for "The number of a journal
+        // or the volume/number of a book in a series".  However, it is also
+        // used for patent entries as "the number or record token of a patent
+        // or patent request", and is also listed as an optional field for
+        // report, manual, and dataset, where it fits the use for patent.
+        // Hayagriva uses "issue" for the journal/book sense of biblatex's number,
+        // and "serial-number" for the record number / token sense.
         if let Some(number) = map_res(entry.number())?.map(|d| d.into()) {
             if let Some(parent) = book(&mut item, parent) {
                 parent.set_issue(number);
@@ -566,6 +573,11 @@ impl TryFrom<&tex::Entry> for Entry {
             item.set_abstract_(abstract_.into())
         }
 
+        // BibLaTeX describes "type" as "The type of a manual, patent, report, or thesis.
+        // This field may also be useful for the custom types listed in § 2.1.3."
+        // Hayagriva uses "genre" for 'Type, class, or subtype of the item (e.g.
+        // "Doctoral dissertation" for a PhD thesis; "NIH Publication" for an NIH
+        // technical report)'
         if let Some(type_) = map_res(entry.type_())? {
             item.set_genre(type_.into());
         }
--- a/src/interop.rs
+++ b/src/interop.rs
@@ -397,7 +397,15 @@ impl TryFrom<&tex::Entry> for Entry {
             if let Some(parent) = book(&mut item, parent) {
                 parent.set_issue(number);
             } else {
-                item.set_issue(number);
+                match item.entry_type {
+                    EntryType::Report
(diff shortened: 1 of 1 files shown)


cgevans:
Test failures appear unrelated here.

cgevans:
Is there anything else that should be done on this for it to be reviewed?

PgBiel:
Sounds good. I've confirmed your choices by taking a look around `apa.csl`[^1], and they appear to match in principle. Thank you!

> I wonder whether non-default styles also use it for other types, eg, Manuscript, Thesis, etc?

At least `apa.csl` seems to accept `number` (hayagriva's `serial`) for `thesis` and `manuscript`, however `ieee.csl`[^2] uses `issue` instead. Maybe we can keep your choices for now as they are less ambiguous and change as needed in the future.

[^1]: https://github.com/citation-style-language/styles/blob/28ee0a509ba772e44d3afbfe7811b7c1bad5ba14/apa.csl
[^2]: https://github.com/citation-style-language/styles/blob/28ee0a509ba772e44d3afbfe7811b7c1bad5ba14/ieee.csl

src/interop.rs
```diff
@@ -558,6 +566,10 @@ impl TryFrom<&tex::Entry> for Entry {
             item.set_abstract_(abstract_.into())
         }
 
+        if let Some(type_) = map_res(entry.type_())? {
+            item.set_genre(type_.into());
```
PgBiel, line 570: I know this isn't really done elsewhere, but do you think you could add a comment here comparing the hayagriva and biblatex docs as you did in the PR description? The same could be done for the other change as well.

PgBiel:
Thank you :blue_heart:

PR #298, opened 2025-03-30
Add locator support to alphanumeric style

### LINKED ISSUE #48: Alphanumerical citation style ignores supplement
From https://github.com/typst/typst/issues/1330.

[CitationStyle::citation for Alphanumeric](https://github.com/typst/hayagriva/blob/a4150683894a51cc57d704c8a81ffac3e5a9532a/src/style/mod.rs#L1077-L1125
) doesn't inspect `atomic.supplement` inside the `for atomic in parts` loop.

Maybe it simply needs something like:

```diff
+           if let Some(supplement) = atomic.supplement {
+              res += ", ";
+              res += supplement;
+           }
            items.push(res);
```

…but maybe it's more complicated than that.

Eloitor:
Fixes #48


styles/alphanumeric.csl
```diff
@@ -22,8 +22,11 @@
       <key variable="issued" />
     </sort>
     <layout prefix="[" suffix="]" delimiter=", ">
-      <text variable="citation-label" />
-      <text variable="year-suffix" />
+      <group delimiter=", ">
+        <text variable="citation-label"/>
+        <text variable="year-suffix"/>
+        <text variable="locator" prefix=" " />
```
PgBiel, line 28: Is there a reason for the nested group with the same delimiter and this prefix? What is your intended output?

Of note, `@citation[abc]`  produced `[Cit25a, abc]` in Typst 0.8.0, before the regression. We should probably stick to that.

PgBiel (changes requested):
Thanks!

Can you add a test like this one? https://github.com/typst/hayagriva/blob/0b32dca41966f267cbd4efbfe42fbf8da3e5e7a3/src/csl/mod.rs#L2996

Eloitor:
what should be the output of `@first[abc] @second[def]`?

Eloitor:
I attempted to add a test, but I don't know if it is possible (or necessary)  to specify create a location in a BibliographyRequest...
```
       let finished = driver.finish(BibliographyRequest {
            style: &alphanumeric,
            locale: None,
            locale_files: &[],
        });
```

PgBiel:
Whoops, pushed the wrong commit to your branch, sorry about that. I'll merge in a separate PR of my own

PgBiel:
(By the way, please use a branch other than your fork's `main` next time. This was totally my fault here though. :joy:)

PgBiel:
> I attempted to add a test, but I don't know if it is possible (or necessary) to specify create a location in a BibliographyRequest...

The problem here was that you needed to update the archive (alphanumeric is also archived under `archive/styles` and that's the file used for the styles, not the csl file). That can be done with

```sh
HAYAGRIVA_ARCHIVER_UPDATE=1 cargo test --test archiver -- --ignored
```

I've done it for you in the new PR though, so don't worry.

PR #299, opened 2025-04-03
Add ID to `LocatorPayload::Transparent`

wkordalski:
This way we can pass information about transparent locator payloads being the same or different,
and thanks to it generate correct `ibid` vs `ibid-with-locator` states.

### DIFF
--- a/src/csl/mod.rs
+++ b/src/csl/mod.rs
@@ -2251,7 +2251,11 @@ pub enum LocatorPayload<'a> {
     Str(&'a str),
     /// An element with the original index of the locator will be yielded. The
     /// consumer can then recognize this and replace it with their own content.
-    Transparent,
+    ///
+    /// Each `Transparent` locator has associated numeric ID. Transparent
+    /// locators with the same ID are considered identical, and with different
+    /// ID are considered different.
+    Transparent(u32),
 }
 
 impl<'a, T: EntryLike> Context<'a, T> {
--- a/src/csl/taxonomy.rs
+++ b/src/csl/taxonomy.rs
@@ -64,7 +64,7 @@ impl<'a, T: EntryLike> InstanceContext<'a, T> {
                         .map(|n| MaybeTyped::Typed(Cow::Owned(n)))
                         .unwrap_or_else(|_| MaybeTyped::String(l.to_owned())),
                 )),
-                LocatorPayload::Transparent => Some(NumberVariableResult::Transparent(
+                LocatorPayload::Transparent(_) => Some(NumberVariableResult::Transparent(
                     self.cite_props.certain.initial_idx,
                 )),
             },


wkordalski:
@PgBiel Can I have your review?

src/csl/mod.rs
```diff
@@ -2251,7 +2251,11 @@ pub enum LocatorPayload<'a> {
     Str(&'a str),
     /// An element with the original index of the locator will be yielded. The
     /// consumer can then recognize this and replace it with their own content.
-    Transparent,
+    ///
+    /// Each `Transparent` locator has associated numeric ID. Transparent
+    /// locators with the same ID are considered identical, and with different
+    /// ID are considered different.
+    Transparent(u64),
```
PgBiel, line 2258: If the intention is to use a hash here, we should use the same type used by Typst for hashes.

```suggestion
    Transparent(u128),
```

PgBiel (commented):
After discussing with the team, we understood that these changes are directly tied to the Typst side of the fix, so we think you can keep this PR open, but in the meantime open a PR in the Typst side pointing to your hayagriva fork (with this PR's changes) showing that this PR  can indeed be used to fix the original issue, e.g. by passing a content hash as the id and then showing in a test that it produces the expected behavior for ibid. Do you think you could do that? :)

wkordalski:
I've created draft PR to Typst: https://github.com/typst/typst/pull/6171

It turned out that two text-only contents with the same text (but in different locations) generate different hashes (thus hashing supplement content is not a good idea).

The PR uses IDs generated from `.plain_text()` of supplements. We can also use hash of plain text, if you like.
This is not perfect solution, but user can use `#text(size: 0pt)[identifier]` trick to mark supplements as different if needed (this workaround is not possible on current main branch version).

I've also created a repository with a "test": https://github.com/wkordalski/typst-hayagriva-ibid
It contains all source files and output generated by Typst from main branch and from the draft PR I've created.
In the next few days I will try to figure out how to write tests in Typst way and rewrite the "test repository" to actual Typst test.

src/csl/mod.rs
PgBiel: On second thought, let's try this approach so our hands aren't tied (since we already have a lifetime parameter anyway):

```suggestion
    Transparent(&'a dyn Eq),
```

wkordalski: Unfortunately, [`Eq` is not dyn compatible](https://doc.rust-lang.org/std/cmp/trait.Eq.html#dyn-compatibility) (think about comparing `&dyn Eq` generated from two different types).

We can introduce generic parameter: `LocatorPayload<'a, T: Eq = ()>`, but I don't think that introducing more generic parameters is the way to go.

We can use `u128` if you want, but I think that we will never need more than `u64`.

Drodt: I would consider a generic parameter with default ‘u64‘  that must implement Copy. Then we know it is reasonably cheap.

wkordalski: I've made locator payload type generic (https://github.com/typst/hayagriva/pull/299/commits/11e1095d4b445a03544038b836046d52bba0d17a).

I don't like adding the transparent locator payload type across the whole hayagriva codebase and would rather stick to `u64`. But if you prefer the generic version, you can merge it.

We can additionally add the default type for the generic parameter, but it is not necessary. Typst doesn't need any change to handle the generic parameter — inference does a great job.

wkordalski:
Current status:
- [Typst part](https://github.com/typst/typst/pull/6171) is ready. It only waits (therefore draft status of PR) for merging this (Hayagriva) PR — and then — for next release of Hayagriva, so that we can remove `[patch.crates-io]` section that overrides Hayagriva version.
- We can discuss whether using `plain_text()` for supplement discrimination is the right choice. I think it is much better than current version and the one that introduces the least complexity.
- On Hayagriva side, we only need to decide whether we want generics or not.
  - Yes → we can add default parameters and merge the PR.
  - No → we can remove last three commits and merge the PR.

PgBiel:
Thank you. The Typst side is currently blocked until further discussion with the main maintainer, who is currently busy with other tasks, so we will have to wait a bit. But I hope to give you further feedback soon.

PgBiel:
Hello, we're currently cleaning the Typst PR backlog. Hope to give you a response soon.

laurmaedje:
> Unfortunately, [Eq is not dyn compatible](https://doc.rust-lang.org/std/cmp/trait.Eq.html#dyn-compatibility) (think about comparing &dyn Eq generated from two different types).
> 
> We can introduce generic parameter: LocatorPayload<'a, T: Eq = ()>, but I don't think that introducing more generic parameters is the way to go.

I would suggest creating a custom type that wraps a `PartialEq` type and encapsulates the trait object. By using `Arc`, we can also keep it cheap to clone without needing a `Clone` impl on the underlying type. It would look like this:

<details>
  <summary>Implementation</summary>

  ```rust
  use std::any::Any;
  use std::fmt::{self, Debug};
  use std::sync::Arc;
  
  #[derive(Clone)]
  pub struct TransparentLocator(Arc<dyn Bounds>);
  
  impl TransparentLocator {
      pub fn new<T>(value: T) -> Self
      where
          T: PartialEq + 'static,
      {
          Self(Arc::new(value))
      }
  }
  
  impl Debug for TransparentLocator {
      fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
          f.pad("TransparentLocator(..)")
      }
  }
  
  impl PartialEq for TransparentLocator {
      fn eq(&self, other: &Self) -> bool {
          self.0.dyn_eq(&*other.0)
      }
  }
  
  trait Bounds: 'static {
      fn as_any(&self) -> &dyn Any;
      fn dyn_eq(&self, other: &dyn Bounds) -> bool;
  }
  
  impl<T> Bounds for T
  where
      T: PartialEq + 'static,
  {
      fn as_any(&self) -> &dyn Any {
          self
      }
  
      fn dyn_eq(&self, other: &dyn Bounds) -> bool {
          let Some(other) = other.as_any().downcast_ref::<Self>() else {
              return false;
          };
          self == other
      }
  }
  
  fn main() {
      let a = TransparentLocator::new("hello");
      let b = TransparentLocator::new("hello");
      let c = TransparentLocator::new("world");
      let d = TransparentLocator::new(5);
      assert_eq!(a, b);
      assert_ne!(a, c);
      assert_ne!(a, d);
  }
  ```

</details>

wkordalski:
> I would suggest creating a custom type that wraps a `PartialEq` type and encapsulates the trait object. By using `Arc`, we can also keep it cheap to clone without needing a `Clone` impl on the underlying type.

That's also a solution.

Would you like to have `TransparentLocator: Eq` or not?
`LocatorPayload` and some other types containing `LocatorPayload` are deriving `Eq` impl.

We can remove the `#[derive(Eq)]` as they are not used in Hayagriva (and then we also can remove `#[derive(Hash)]` as it usually makes sense only when type is also `Eq`).

Or we can wrap types that are `Eq` and make the whole `TransparentLocator` implement `Eq`.


What is the policy of Typst on adding small dependencies? There's `dyn_eq` crate that does exactly what suggested implementation does.

PgBiel (commented):
Thank you!

src/csl/mod.rs
```diff
@@ -2242,17 +2245,48 @@ impl IbidState {
...
+
+impl PartialEq for TransparentLocator {
+    fn eq(&self, other: &Self) -> bool {
+        TransparentLocatorPayload::dyn_eq(self.0.as_ref(), other.0.as_ref())
+    }
+}
+
+trait TransparentLocatorPayload: Any {
+    fn dyn_eq(&self, other: &dyn TransparentLocatorPayload) -> bool;
+}
```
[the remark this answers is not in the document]
PgBiel: Oh okay!

PgBiel (approved):
It's a bit unfortunate that we lose Eq and Hash everywhere, but I suppose that isn't very useful for those particular structs.

Otherwise, the implementation looks fine.
We can merge after the necessary changes are applied to https://github.com/typst/typst/pull/6171 and verified to work with this diff (you can temporarily point the hayagriva dep to this PR's branch there).

Notably, we'll now be able to use the supplement as Content directly rather than `plain_text()`.

wkordalski:
> Notably, we'll now be able to use the supplement as Content directly rather than `plain_text()`.

Earlier:
> It turned out that two text-only contents with the same text (but in different locations) generate different hashes (thus hashing supplement content is not a good idea).

This means also that for two text-only contents with the same text (but in different locations) will be different according to PartialEq.

I will check this — maybe something changed.

laurmaedje:
I think some problems with CI were fixed, so maybe rebasing this is enough to get CI to be green?

wkordalski:
Now, after rebase the Hayagriva CI passes. However [4 of the Typst tests still fails](https://github.com/typst/typst/actions/runs/17360476377).

I've created [another PR](https://github.com/typst/typst/pull/6842) to Typst that upgrades Hayagriva version to the one from main branch (probably not to be merged). It also fails with the same errors. This proves that it's not this PR that introduces theses errors.

Therefore we probably can safely merge this PR.

PgBiel:
Yeah, the errors you are facing are due to this issue https://github.com/typst/hayagriva/issues/371. We don't know yet what caused this change. Sorry for the trouble.

Drodt (approved):
Looks good! Thank you for finally fixing that feature!

PR #313, opened 2025-05-06
Fix name sorting when using `form=count`

Drodt:
For some reason, name forms where disabled during sorting, which let to a number of sorting tests failing when they rely on the `count` name form.

This PR adds 7 tests to citeproc-pass.


PgBiel:
The reason is likely the spec's wording regarding sorting by variables, where the default name form for name variables is specified[^1]:

> names: Name variables called via the variable attribute (e.g. <key variable="author"/>) are returned as a name list string, with the cs:name attributes form set to “long”, and name-as-sort-order set to “all”.

However, setting a different attribute form is explicitly allowed when sorting by a macro that produces names (which is what happens in the tests that now pass):

> There are four advantages in using the same macro for rendering and sorting, instead of sorting directly on the name variable. (...) Third, names can be sorted by just the surname (using a macro for which the `form` attribute on `cs:name` is set to "short"). Fourth, it is possible to sort by the number of names in a name list, by calling a macro for which the `form` attribute on `cs:name` is set to "count".

Do you think your change is enough to ensure variables are always rendered with the long form? (When could this not happen?)

[^1]: https://docs.citationstyles.org/en/stable/specification.html#sorting-variables

Drodt:
> Do you think your change is enough to ensure name variable keys (outside macros) are always rendered with the long form? (When could this not happen?)

Yes. If the name variable is directly used in `<sort>`, then hayagriva already overwrites the name form to long and uses `Person::csl_cmp` instead of normal rendering. From what I see, the check during the rendering I removed was unnecessary for that behavior.

PR #315, opened 2025-05-06
Apply Clippy Fixes

Drodt:
Applies the clippy fixes for newer versions of clippy. Improves code quality slightly. Bumps the Rust version for the CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@1.86.0
+      - uses: dtolnay/rust-toolchain@1.85.0
         with:
           components: clippy, rustfmt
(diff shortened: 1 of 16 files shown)


.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@1.83.0
+      - uses: dtolnay/rust-toolchain@1.86.0
```
PgBiel, line 22: We shouldn't bump ci to a Rust version that is ahead of Typst at the moment.

PgBiel (changes requested):
Changes look good. Thanks!

.github/workflows/ci.yml
Drodt: Fair enough. Changed to 1.85, which typst uses

PgBiel (approved):
Thanks!

PR #126, opened 2024-01-11
Multimodal bibliography entries

nrydanov:

PR #317, opened 2025-05-07
Map biblatex language to hayagriva

### LINKED ISSUE #292: sort by language not work
It seems that `language` variable was not read. It is blank.

```typst
#cite(<Dong_2022_Can>, form: "prose")

// #[@Dong_2022_Can]
#[@zhangshenglei_2006_jiyu]

#[@zhao_2016_shuiwen]

#[@Xu_2016_Diversity]

#show bibliography: it => {
  set text(size: 11.5pt, font: ("Times New Roman")) // "SimSun"
  set par(spacing: 1.24em, leading: 1em)

  show regex("([^.,:\s]+(?:期刊|杂志|学报|通报|科学|大学学报|研究)[^.,:\s]*)"): match => {
    strong(emph(match.text))
  }
  it
}

*参考文献*
#v(-2.5em)
#bibliography("test.bib", title: "", 
  style:"csl/001gb-t-7714-2015-author-date-bilingual.csl")
```

```xml
  <bibliography entry-spacing="0" et-al-min="10" et-al-use-first="10" hanging-indent="true">
    <sort>
      <!-- <key variable="langid"/> -->
      <key variable="language"/>
      <!-- <key macro="author"/> -->
      <!-- <key macro="issued-year"/> -->
    </sort>
    <!-- 取消这部分注释可以开启 CSL-M 的多语言功能：按照文献的语言输出“et al.”等术语 -->
    <!-- <layout suffix="." locale="en"><text macro="entry-layout"/></layout> -->
    <layout suffix=".">
      <text variable="citation-number" prefix="[" suffix="] "/>
      <text macro="entry-layout"/>
    </layout>
    <!-- <layout suffix=".">
      <text macro="entry-layout"/>
    </layout> -->
  </bibliography>
```

Image

Drodt:
The language field of biblatex was previously ignored, which caused #292. This PR adds the conversion.

If the field is not a valid language identifier, we ignore it. This may be improved in the future.

Close #292


src/interop.rs
```diff
@@ -276,7 +277,10 @@ impl TryFrom<&tex::Entry> for Entry {
             item.add_affiliated_persons((a, PersonRole::Translator));
         }
 
-        // TODO: entry.orig_language into item.language = Some()
+        // If we cannot parse the language, ignore it
+        if let Some(Ok(l)) = map_res(entry.language())?.map(|l| l.parse()) {
+            item.language = Some(l)
```
PgBiel, line 282: We should switch to `item.set_language` for consistency.

src/interop.rs
```diff
@@ -276,7 +277,10 @@ impl TryFrom<&tex::Entry> for Entry {
             item.add_affiliated_persons((a, PersonRole::Translator));
         }
 
-        // TODO: entry.orig_language into item.language = Some()
+        // If we cannot parse the language, ignore it
+        if let Some(Ok(l)) = map_res(entry.language())?.map(|l| l.parse()) {
```
PgBiel, line 281: I'm not sure if this parse call accepts common language names, but I have a hunch that this is just never going to parse for commonly used biblatex files, going by examples from https://tex.stackexchange.com/questions/323441/how-does-langid-field-in-biblatex-differ-from-language-field and https://forums.zotero.org/discussion/110863/biblatex-both-langid-and-language-are-needed from my research. I wonder if there's something we can do about that?

In particular, the biblatex manual has a table of supported languages. Maybe we can have some conversion going on here. (Perhaps even in the `biblatex` library with some enum or using `LanguageIdentifier` directly.)

Not to mention that, per the biblatex manual, this can actually be a list (separated by "and"). Which suggests that the type in the `biblatex` library is wrong; it should be a vector of strings (or language enums) instead of a simple string.

Which is evidently incompatible with CSL's model, which expects a single language per the spec, but we should at least make the explicit choice of either discarding certain languages or simply ignoring bib entries with multiple languages.

PgBiel, line 281: As per the TODO comment, we should probably not ignore `orig_language`. Per the biblatex manual:

> **origlanguage** list (key)
> If the work is a translation, the language(s) of the original work.

Though `origlocation` also exists, as well as `origdate`, and we don't seem to use those anywhere. `origpublisher` isn't even parsed by `biblatex` (perhaps a bug). So I guess this isn't a new problem.

I suppose the correct approach in the hayagriva model would be to have those "orig" fields to be part of a (pre-translation) parent entry? Not sure. We can also just ignore this for now then.

PgBiel (commented):
Thanks! I think there are some things to work out.

PgBiel:
This will conflict with https://github.com/typst/hayagriva/pull/126/, which has some important considerations. We should first check if there will be a conflict with those considerations. The idea of converting languages from biblatex's format has also been mentioned there.

src/interop.rs
Drodt: Opened the very brief #319 for this.

Drodt:
> This will conflict with #126, which has some important considerations. We should first check if there will be a conflict with those considerations. The idea of converting languages from biblatex's format has also been mentioned there.

Is that PR still active? It seems like this PR might be a good first step and #126 is the eventual improvement.

PgBiel:
We don't have to go all the way that #126 did so far of translating dates and other parts of the entry based on the language, but we should at least make sure interop is correct. To this end, we might need changes on `biblatex` to solve https://github.com/typst/hayagriva/pull/317#discussion_r2087729074.

src/interop.rs
PgBiel: Let's ignore it for now.

src/interop.rs
```diff
@@ -276,7 +277,27 @@ impl TryFrom<&tex::Entry> for Entry {
...
+            match l {
+                PermissiveType::Typed(lang) => {
+                    item.set_language((*lang).into());
+                }
+                PermissiveType::Chunks(spanneds) => {
+                    let chunked: ChunkedString = (spanneds as &[Spanned<Chunk>]).into();
+                    if let Ok(l) = chunked.to_str().parse() {
+                        item.set_language(l);
+                    }
+                }
```
PgBiel, line 298: i think we should refrain from parsing invalid languages from the time being, even if they are valid from a Unicode perspective, to ensure we will be respecting the biblatex standard. We can revise this in the future if needed.

src/interop.rs
```diff
@@ -276,7 +277,27 @@ impl TryFrom<&tex::Entry> for Entry {
...
 
-        // TODO: entry.orig_language into item.language = Some()
+        // Take the first language or the langid.
+        let lang_res = entry.language();
+        let langid_res = entry.langid().ok();
+        // If we cannot parse the language, ignore it
+        if let Some(l) = map_res(lang_res)?
+            .as_ref()
+            .and_then(|l| l.first())
+            .or(langid_res.as_ref())
```
PgBiel, line 287: I'm a bit unsure regarding `language` vs `langid` here. If I understood correctly from the `biblatex` manual and from my research, `language` indicates the language(s) in which the original work is available, whereas `langid` indicates the language that should be used for typesetting purposes, e.g. for hyphenation.

It seems less clear how to map this to CSL. CSL has the notion of locales which specify the values of `term`. OTOH, the `language` field seems to be used to control how title case conversion of strings is performed. I think the original language of an entry should probably not affect the `term` used for example, it could be inconsistent to add foreign words to your document's bib based on the origin of the work, though I couldn't find anything in CSL mentioning that `language` affects a `term` so maybe that is not a concern. And changing the output of title casing seems sensible, but I think this suggests `langid` is more appropriate since that's more or less controlling the typesetting of the entry.

On the other hand, using langid means styles would not be able to display the language(s) of a work, which is what the `biblatex` field `language` is for, since they would not receive this information anymore. This is a bit concerning, but I think this is more of a deficiency in the CSL standard. But I'm still not sure, and using both langid and language as you are doing seems like a good compromise for now, though maybe `langid` should be prioritized and not the other way around.

This still needs a bit of research before proceeding.

Cargo.toml
```diff
@@ -24,11 +24,13 @@ paste = "1.0.14"
 serde = { version = "1", features = ["derive"] }
 serde_yaml = "0.9.25"
 thiserror = "1.0.48"
-unic-langid = { version = "0.9.0", features = ["serde"] }
+unic-langid = { version = "0.9.6", features = ["serde"] }
 unicode-segmentation = "1.6.0"
 unscanny = "0.1.0"
 url = { version = "2.4", features = ["serde"] }
-biblatex = { version = "0.10.0", optional = true }
+biblatex = { git = "https://github.com/DerDrodt/biblatex.git", branch = "lang", features = [
```
PgBiel, line 31: We can do this now I think.
```suggestion
biblatex = { git = "https://github.com/typst/biblatex.git", features = [
```

PgBiel (commented):
Thanks!

src/interop.rs
Drodt: The biblatex documentation mentions that key values, like a language, are printed as is if they are not recognized. I can test it out if that is correct. If so, should we then handle chunks here?

src/interop.rs
Drodt: My impression was that ’language’ is more appropriate because hayagriva‘s language field is for the language the entry was written in. I don’t see it affecting the rendering language but I will investigate

PgBiel: Let's keep it like this for now then.

PgBiel:
Yeah you're right, a bit of an oopsie on my part. Thanks!

PR #338, opened 2025-06-30
Update archive

Drodt:


PgBiel:
> Tests fail because the English locale was changed and the tests have not been updated

I guess we should wait a bit then

PgBiel:
Seems like we can proceed: https://github.com/citation-style-language/test-suite/pull/75

Drodt:
We still have to wait for 
citation-style-language/test-suite#76

Drodt:
Now we're ready.

PR #350, opened 2025-07-24
Update archives & Citationberg

Drodt:
Uses typst/citationberg#25


Drodt:
I don't know why the test is failing

PgBiel:
Thanks!

PgBiel:
It seems that this PR broke Typst's tests as `chicago-notes` is gone. I wonder why that is? (cf. https://github.com/typst/hayagriva/pull/299#issuecomment-3196549648)

We should add it back somehow.

PR #364, opened 2025-08-28
Use locale terms for AD/BC

Drodt:
Previously it was hard coded to English (without leading spaces).

### DIFF
--- a/src/types/time.rs
+++ b/src/types/time.rs
@@ -252,17 +252,17 @@ impl Date {
         ad_prefix: bool,
     ) -> String {
         let np_postfix = match (secular, periods) {
-            (true, false) => " BCE",
-            (true, true) => " B.C.E.",
-            (false, false) => " BC",
-            (false, true) => " B.C.",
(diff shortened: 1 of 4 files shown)


src/types/time.rs
```diff
@@ -252,17 +252,17 @@ impl Date {
...
         let positive_dn = match (periods, ad_prefix) {
-            (true, false) => "C.E.",
-            (false, false) => "CE",
-            (true, true) => "AD",
-            (false, true) => "A.D.",
+            (true, false) => " C.E.",
+            (false, false) => " CE",
+            (true, true) => " A.D.",
+            (false, true) => " AD",
         };
```
PgBiel, line 266: Other than the changes in periods, were those changes really necessary? We seem to already be using spaces if necessary when writing below

PgBiel (commented):
Thanks!

PgBiel:
Big thanks! That was an annoying one to fix haha.

PR #367, opened 2025-08-28
Fix year suffix collapse

Drodt:
Fixes collapsing of year suffixes.

Adds 

1. bugreports_BadDelimiterBeforeCollapse
2. collapse_YearSuffixCollapseNoRange
3. date_LopsidedDataYearSuffixCollapse
4. date_YearSuffixDelimiter
5. disambiguate_YearSuffixMixedDates
6. punctuation_DefaultYearSuffixDelimiter

to the passed tests


src/csl/mod.rs
```diff
@@ -969,14 +970,29 @@ fn collapse_items<'a, T: EntryLike>(cite: &mut SpeculativeCiteRender<'a, '_, T>)
...
         }
-        Some(Collapse::Year | Collapse::YearSuffix | Collapse::YearSuffixRanged) => {
+        Some(
+            c @ (Collapse::Year | Collapse::YearSuffix | Collapse::YearSuffixRanged),
+        ) => {
             // Index of where the current group started and the group we are
             // currently in.
             let mut group_idx: Option<(usize, usize)> = None;
             for i in 0..cite.items.len() {
+                eprintln!("{:?}", cite.items[i].cite_props);
```
PgBiel, line 989: You forgot this here

src/csl/mod.rs
```diff
@@ -969,14 +970,29 @@ fn collapse_items<'a, T: EntryLike>(cite: &mut SpeculativeCiteRender<'a, '_, T>)
...
                 match group_idx {
                     // This is our group.
                     Some((_, idx)) if Some(idx) == cite.items[i].group_idx => {
+                        if c == Collapse::YearSuffix {
+                            match cite.items[i].cite_props.speculative.disambiguation {
+                                DisambiguateState::YearSuffix(s) if s > 0 => {
```
PgBiel, line 995: This construct can be replaced with a single `if` thanks to `let Some`, such as
```rs
if c == Collapse::YearSuffix && let DisambiguateState::YearSuffix(s) = cite.items[i].cite_props.speculative.disambiguation && s > 0 {
```

PgBiel (commented):
Thanks! I still want to take a look at the logic, but here are some immediate comments.

src/csl/mod.rs
Drodt: Thanks!

src/csl/mod.rs
Drodt: Right, forgot that's stable now.

PgBiel (approved):
Thanks!

PR #372, opened 2025-09-01
Fix missing styles; future proof archiver; update styles

Drodt:
- Add a check to the archiver that no expected styles are missed
- Updates styles, adapted to new chicago structure in the CSL repo
- Removed turabian styles; they are now dependent


tests/archiver.rs
```diff
@@ -587,8 +605,8 @@ const OVERRIDES: [Override; 19] = [
         "china-national-standard-gb-t-7714-2015-author-date",
         "gb-7714-2015-author-date",
     ),
-    Override::first("chicago-fullnote-bibliography", "chicago-fullnotes"),
-    Override::first("chicago-note-bibliography", "chicago-notes"),
+    Override::first("chicago-notes-bibliography", "chicago-notes"),
+    Override::first("chicago-shortened-notes-bibliography", "chicago-shortened-notes"),
```
PgBiel, line 609: Wonder if we can add fullnotes back as an alias to soften the impact of this change?

PgBiel (commented):
Thanks!

tests/archiver.rs
```diff
@@ -580,15 +598,16 @@ impl Override {
...
     Override::alias("apa", "american-psychological-association", &["apa"]),
     Override::alias("bmj", "british-medical-journal", &["bmj"]),
     Override::first(
         "china-national-standard-gb-t-7714-2015-author-date",
         "gb-7714-2015-author-date",
     ),
-    Override::first("chicago-fullnote-bibliography", "chicago-fullnotes"),
-    Override::first("chicago-note-bibliography", "chicago-notes"),
+    Override::first("chicago-notes-bibliography", "chicago-notes"),
+    Override::first("chicago-shortened-notes-bibliography", "chicago-fullnotes"),
```
PgBiel, line 609: This doesn't seem right. `chicago-fullnotes` should alias `chicago-notes`. I feel like you will need to use `Override::alias` for this like in the apa example.

In addition, shortened-notes should be available on its own. (Fixing this line should be enough.)

PgBiel (commented):
Thanks. I was experimenting a bit further on the Typst side, and I noticed a problem:

PR #361, opened 2025-08-20
Stop `chapter` from breaking BibTeX-based entries

fpdotmonkey:
`chapter` in BibLaTeX represents "a chapter or section or any other unit of a work" according to its manual [1][biblatex].  In the BibTeX manual, it's said to be "a chapter (or section or whatever) number" [2][bibtex].

I understand this to mean that in Bib(La)TeX land, `chapter` is something akin to `pages`, an indication to the curious reader where to look to find the source.  Suffice to say, the current state where near all other bibliographic information is thrown away to make way for an `EntryType::Chapter` misunderstands the semantics of `chapter` (see

I believe to achieve the same semantics as Bib(La)TeX, a new field would need to be added to `Entry` to represent a chapter/section number.  I'll leave that to someone else to figure out.  Until then, this change prevents `chapter` from borking your bibliography entry.

Closes #357

[biblatex]: https://www.texlive.info/CTAN/macros/latex/contrib/biblatex/doc/biblatex.pdf
[bibtex]: https://www.nic.funet.fi/pub/TeX/CTAN/biblio/bibtex/base/btxdoc.pdf

PR #383, opened 2025-09-15
Add `chapter` field

PgBiel:
This is a follow-up to #361.

We decided to simply remove the automatic conversion of BibTeX book/incollection entry type to chapter for now, as that yields more useful results in general, given the chapter entry type is less supported. Usually, the chapter field is just adjacent information, whereas `@InBook` would be better suited for the `chapter` entry type in CSL (we already perform this conversion).

CSL also has a `chapter-number` field which seems to have the same purpose as the Bib(La)TeX chapter field, which is to simply add extra information about where to find this citation. This PR creates a field for it so we can then map that BibTeX field to it.

However, I'm a bit unsure about adding something like `chapter: 3` for two reasons:
1. This might discourage people from using the chapter entry type altogether. Although it does feel a bit weird on its own.
2. Currently, `entry: chapter` uses `serial-number: 3` to indicate that chapter 3 is being cited. It feels then that `chapter: 3` would have to map to `number` (i.e. function in the same way) for that entry type, as it feels like it's what the user would pick to indicate the chapter being cited.

An alternative is to use `serial-number: { chapter: ... }` which, I guess, would be good enough to discourage users from using it without checking what it does in the docs or something. But it also feels very unnatural.
Paging @reknih for opinions.

### DIFF
--- a/src/csl/taxonomy.rs
+++ b/src/csl/taxonomy.rs
@@ -146,16 +146,19 @@ impl EntryLike for Entry {
     ) -> Option<MaybeTyped<Cow<'_, Numeric>>> {
         match variable {
             NumberVariable::ChapterNumber => self
-                .bound_select(
-                    &select!(
-                        (("e":Anthos) > ("p":Anthology)) |
-                        (("e":*) > ("p":Reference)) |
-                        (("e":Article) > ("p":Proceedings)) |
-                        (("e":*) > ("p":Book))
-                    ),
-                    "e",
-                )
-                .and_then(Entry::volume)
+                .chapter()
+                .or_else(|| {
+                    self.bound_select(
+                        &select!(
+                            (("e":Anthos) > ("p":Anthology)) |
+                            (("e":*) > ("p":Reference)) |
+                            (("e":Article) > ("p":Proceedings)) |
+                            (("e":*) > ("p":Book))
+                        ),
+                        "e",
+                    )
+                    .and_then(Entry::volume)
+                })
                 .map(MaybeTyped::to_cow),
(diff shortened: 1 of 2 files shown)


fpdotmonkey:
So the ultimate problem that's we're trying to solve is allowing a user to have the chapter number in a reference when a BibTeX entry has a `chapter` field.  Importantly, this should render similar output to biblatex/biber given the same input.

From my study on this, we need to rectify four layers of semantics to achieve this.

1. BibLaTeX `chapter` field
2. Hayagriva `chapter (to bikeshed, I'd suggest `chapter-number` to avoid confusion)
3. CSL `chapter-number` field (probably)
4. Each style's interpretation of when to render the chapter number

The first three are straightforwardly rectified by this PR.  I've tested it and it works insofar as you can see with a debugger that the entry object has the chapter that you'd expect.  It does not work in that with popular styles like IEEE, the chapter number doesn't render.

This is actually a bit tricky.  In `archive/styles/ieee.cbor`, `chapter-number` is only used with `chapter` entry types.  On the other hand, `archive/styles/chicago-author-date.cbor` uses it for `personal_communication`, `entry`, and legal stuff.  It doesn't seem that CSL themes are terribly flexible nor consistent about when you can use `chapter-number`.

Does rectifying this require submitting changes to the upstream style vendors?

fpdotmonkey:
> Did you actually read the CBOR (compressed) data directly? I'm impressed haha.

No, I parsed the data to JSON and formatted it haha

```sh
cbor2json.rb archive/styles/ieee.cbor | jq -C | less -R
```

PgBiel:
Thanks a lot!

Issue #325, opened 2025-05-29
CSL `<date>` element's `delimiter` attribute is ignored


kristoferssolo:
The `delimiter` attribute on the CSL `<date>` element, appears to be ignored by Hayagriva. This leads to incorrectly formatted dates for locales that rely on this attribute for spacing between date parts.

### CSL Specification
According to the CSL documentation on [Date Formats](https://docs.citationstyles.org/en/stable/translating-locale-files.html#date-formats), the `<date>` element accepts a `delimiter` attribute to specify the string that separates rendered date-parts.

### Example from Latvian Locale (lv-LV)

Consider the following CSL definition for a text date format:

```xml
<date form="text" delimiter=" ">
  <date-part name="year" suffix=". gada"/>
  <date-part name="day" form="numeric" suffix="."/>
  <date-part name="month"/>
</date>
```

#### Expected Behavior
For a date like `2025-05-29`, the `delimiter=" "` should insert a space between "gada" and the day, and between the day's number and the month.
The expected output should be: `2025. gada 29. maijā`

#### Actual Behavior
`2025. gada29.maijā` (note the missing spaces).

PgBiel:
This behavior appears to be specific to localized date formats. The spec indicates affixes for `date` are ignored for localized date formats, and a comment in citationberg indicates this also applies to `delimiter`. Indeed, the spec appears to include e.g. spaces between items as part of prefixes and suffixes in the format. I couldn't find a more explicit confirmation of this though.

PgBiel:
Fixed by #385

PR #385, opened 2025-09-15
Fix date delimiter handling for localized date formats (#325)

kristoferssolo:
Fixes a bug where the delimiter from a localized CSL date format (locale file) was ignored in favor of the cs:date element’s own delimiter. As a result some locales (e.g. Latvian) produced merged date parts like "2025. gada15.septembrī" instead of the expected "2025. gada 15. septembrī".

### DIFF
--- a/src/csl/rendering/mod.rs
+++ b/src/csl/rendering/mod.rs
@@ -674,10 +674,20 @@ impl RenderCsl for citationberg::Date {
         // TODO: Date ranges
         let mut last_was_empty = true;
 
-        // Localized (base) delimiter takes precedence over the cs:date element's delimiter attribute.
-        let chosen_delim = base
-            .and_then(|b| b.delimiter.as_deref())
-            .or(self.delimiter.as_deref());
+        // Localized (base) delimiter takes precedence over the cs:date
+        // element's delimiter attribute, which should be ignored if a locale's
+        // date form is used, according to the CSL 1.0.2 spec (under "Style
+        // Behavior"):
+        //
+        // "Delimiter
+        //
+        // The delimiter attribute, whose value delimits non-empty pieces of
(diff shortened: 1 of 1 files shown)


PgBiel:
Your test wasn't actually testing it, as it wasn't using a localized date format: that only applies when either `form="text"` or `form="numeric"` is used, and then the locale fully takes over formatting.

Interestingly, the spec says 

> [Delimiter](https://docs.citationstyles.org/en/stable/specification.html#toc-entry-70)
>
> The delimiter attribute, whose value delimits non-empty pieces of output, may be set on cs:date (delimiting the date-parts; delimiter is not allowed when cs:date calls a localized date format) (...)

Meaning the current behavior of falling back to the overridden delimiter is wrong. I'm not sure what the ideal behavior would be to satisfy "not allowed", but my hunch is that we would simply not use a delimiter at all in that case (locale specifies no delimiter), even if overridden.

Though I wonder if any styles currently rely on this behavior. I guess, if they did, then they shouldn't be using `form="..."` anyway.

PgBiel:
Using `rg "form=\"(text|numeric)\".*\\sdelimiter|\\sdelimiter=.*form=\"(text|numeric)\"" -l | wc -l` on the official styles repo (https://github.com/citation-style-language/styles/commit/c729cc28a78e9fb6b89e8873c5d64e5a7f2536ca) I could find 108 styles overriding the delimiter while also using `form="text"` or `form="numeric"`.

I wonder if they rely on this or not...

Maybe we should play it safe and keep the fallback anyway... but I feel kinda bad about ignoring the spec here. Or maybe my interpretation is unnecessarily strict.

PgBiel:
Yeah, I think that is the most spec-accurate interpretation, and is basically what I've implemented now.

Thank you!

PR #390, opened 2025-09-22
Fix sorting for empty sort values

Drodt:
Entries where the sort value does not exist should always be put to the end.

Adds the following tests to citeproc-pass:

1. date_NonexistentSortReverseBibliography
2. date_NonexistentSortReverseCitation
3. date_SortEmptyDatesBibliography
4. date_SortEmptyDatesCitation
5. sort_StatusFieldAscending

### DIFF
--- a/src/csl/sort.rs
+++ b/src/csl/sort.rs
@@ -140,6 +140,7 @@ impl StyleContext<'_> {
             }
         };
 
+        // Per CSL 1.0.2 spec (https://docs.citationstyles.org/en/v1.0.2/specification.html):
         // Entries with empty values are always displayed at the end, even with reversed order.
         // So we always reverse the order if either entry is empty.
         if empty_value || key.sort_direction() == SortDirection::Descending {


tests/citeproc.rs
```diff
@@ -434,7 +434,7 @@ impl TestSuiteResults {
 #[ignore]
 fn test_single_file() {
     let locales = locales();
-    let name = "bugreports_ArabicLocale.txt";
+    let name = "date_NonexistentSortReverseBibliography.txt";
```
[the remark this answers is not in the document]
Drodt: Nothing. It's reverted :)

src/csl/sort.rs
```diff
@@ -133,11 +136,13 @@ impl StyleContext<'_> {
                 let a_rendered = render(a, a_idx);
                 let b_rendered = render(b, b_idx);
 
-                a_rendered.cmp(&b_rendered)
+                (a_rendered.cmp(&b_rendered), false)
             }
         };
 
-        if key.sort_direction() == SortDirection::Descending {
+        // Entries with empty values are always displayed at the end, even with reversed order.
```
PgBiel, line 143: Is this behavior in the spec?

Drodt: Yes, see the [sorting section](https://docs.citationstyles.org/en/stable/specification.html?#sorting):

> Items with an empty sort key value are placed at the end of the sort, both for ascending and descending sorts.

PgBiel: Thanks, can you reference it explicitly in the comment? Something like "Per the CSL 1.0.2 spec, § Sorting: ..."

PR #397, opened 2025-09-25
Expose `TransparentLocator` again

PgBiel:
Undo accidental regression from #383 that removed TransparentLocator from public exports (after #299).

Reported by @SillyFreak at https://github.com/typst/hayagriva/pull/383#issuecomment-3335993578 (thank you very much!).


SillyFreak:
Seems to work. Right now I just basically ignore the new feature (https://github.com/SillyFreak/typst-alexandria/pull/22/commits/ffa705ae231126c9e034003665cd9d503932cc2f#diff-8d9574d8f17aa6f384bc6d3c7f19395e4c2434fac4f4b1d1e8994c14cc446b95) but I can compile and no regressions, so as far as Hayagriva's job in this goes, it should be fine.

The better confirmation would be to update https://github.com/typst/typst/pull/6171 to `1796f0e` of course

PR #399, opened 2025-09-26
Fix wrong ibid position, test transparent locators

PgBiel:
According to the CSL spec[^1] (emphasis mine):

> The presence of locators determines which position is assigned:
> 
>  - Preceding cite does not have a locator: if the current cite has a locator, the position of the current cite is “ibid-with-locator”. Otherwise the position is “ibid”.
>  - Preceding cite does have a locator: if the current cite has the same locator, the position of the current cite is “ibid”. If the locator differs the position is “ibid-with-locator”. **If the current cite lacks a locator its only position is “subsequent”.**

Prior logic only checked if the locators differed in any way, assigning "ibid with locator" if so. But that included differing by the previous citation having a locator and the current one not having one, whereas that case should be handled differently per the spec, and the position should be subsequent instead.
(Related: #301, though that fixed a different bug.)

Added tests for transparent locators, which not only verify this behavior has been fixed but also properly test the additions in #299.

[^1]: https://docs.citationstyles.org/en/stable/specification.html#choose

### DIFF
--- a/src/csl/mod.rs
+++ b/src/csl/mod.rs
@@ -2239,7 +2239,8 @@ impl DisambiguateState {
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum IbidState {
-    /// The previous cite referenced another entry.
+    /// The previous cite referenced another entry, or this cite has no locator
+    /// when the previous one did.
     Different,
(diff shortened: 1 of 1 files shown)


reknih (approved):
First I thought it was wrong, but upon closer inspection, the second branch says "This is none, last is some" so it matches the spec.

PR #400, opened 2025-09-26
Don't count year-suffix as a variable access

### LINKED ISSUE #392: Choose not rendered due to possibly incorrent `UsageInfo`
The `date_YearSuffixImplicitWithNoDate` citeproc test is failing. I tried to investigate why but I am stuck.

Assume the following CSL style
```xml
<citation disambiguate-add-year-suffix="true">
    <layout delimiter="; " prefix="(" suffix=")">
      <group delimiter=" ">
        <names variable="author">
          <name/>
        </names>
        <group>
          <choose>
            <if variable="issued">
              <date variable="issued" form="text" date-parts="year"/>
              <text variable="year-suffix"/>
            </if>
            <else>
              <text term="no date" form="short"/>
              <text variable="year-suffix" prefix="-"/>
            </else>
          </choose>
        </group>
      </group>
      <group delimiter=" " prefix=" [" suffix="]">
        <text term="accessed" suffix=": " text-case="capitalize-first"/>
        <date variable="accessed" form="text" date-parts="year-month-day"/>
      </group>
    </layout>
  </citation>
```

And the cited item
```json

{
    "accessed": {
      "date-parts": [
        [
          "1965",
          "6",
          "1"
        ]
      ]
    },
    "author": [
      {
        "family": "Doe",
        "given": "John"
      }
    ],
    "id": "ITEM-1",
    "type": "book"
  }
```

Then, clearly, the `issued` date is missing, so we ought to render `n.d.` (and possibly a suffix). But Hayagriva renders nothing. I tracked it down to this:

The entire group around the choose is not rendered, because the group's `UsageInfo::should_render_group` returns false. The reason is that `has_vars` is true but `has_non_empty_var`, `has_used_macros`, and `has_non_empty_group` are false. That seems correct, but I don't understand why that should suppress rendering.

Since it is not commented, does anyone have any insights @reknih @PgBiel ?

Drodt:
Adds group_ComplexNesting to citeproc-pass. Closes #392.

### DIFF
--- a/src/csl/rendering/mod.rs
+++ b/src/csl/rendering/mod.rs
@@ -169,7 +169,15 @@ impl RenderCsl for citationberg::Text {
         if suppressing {
             ctx.writing.start_suppressing_queried_variables();
         }
-        let targets_variable = matches!(self.target, TextTarget::Variable { .. });
+        let targets_variable = match self.target {
+            // Year suffix is not counted as a variable
+            TextTarget::Variable {
+                var: Variable::Standard(StandardVariable::YearSuffix),
+                ..
+            } => false,
+            TextTarget::Variable { .. } => true,
+            _ => false,
+        };
 
         let Some(target) = lookup_result else {
             return (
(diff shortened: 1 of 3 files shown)


Drodt:
This behavior mirrors citeproc. I found nothing specific in the spec but it kind of makes sense to me. Some style like APA use exactly the pattern of the test in `date_YearSuffixImplicitWithNoDate`, which I hoped to fix (but there is more wrong with that one).

E.g.:
```xml
<text term="no date" form="short"/>
<text variable="year-suffix" prefix="-"/>
```
This is printed if "issued" and "status" don't exist. Without this PR, the "no date" would never render.

PR #409, opened 2025-10-16
Add option for offsetting the citation number

thuiop:
This was (briefly) discussed in the Typst discord; the aim is to allow managing several different bibliographies at once with hayagriva while keeping independent numbering (e.g. the first bibliography will have [1] and [2] and the second one will have [3]). I tried to do it in a fairly minimal way; this should not change the interface for regular usage. I am happy to discuss the change, and add tests or documentation if needed.


laurmaedje:
Thanks!

PR #415, opened 2025-11-06
Do not render empty et al.

Drodt:
If et al. is used and and zero names precede it, we do not render it at all.

Fixes one test.


PgBiel (commented):
Feels like a bit of a contrived use case (omitting all names to append et al sounds weird), but I guess we should still handle it. Thanks!

PR #417, opened 2025-11-07
Improve CSL test parsing

Drodt:
The CSL test fixtures are not formatted as well as Hayagriva assumed. I relaxed the requirements a lot and now the only parsing errors are due to JSON errors (because some tests contain an empty object for some reason) and XML errors due to tafia/quick-xml#674.

7 additional tests are added; 2 of them pass; 3 are skipped:
- bugreports_EnvAndUrb: Skipped, uses `CITATIONS`
- affix_PrefixFullCitationTextOnly: Skipped, uses prefix
- affix_PrefixWithDecorations: Skipped, uses prefix
- date_OtherAlone: Fails, uses literal date
- number_SpacesMakeIsNumericFalse: Passes
- textcase_LocaleUnicode: Fails, we don't give the same error message
- date_YearSuffixImplicitWithNoDateOneOnly: Passes

### DIFF
--- a/tests/citeproc-pass.txt
+++ b/tests/citeproc-pass.txt
@@ -136,6 +136,7 @@ date_SeasonSubstituteInGroup
 date_SortEmptyDatesBibliography
 date_SortEmptyDatesCitation
 date_YearSuffixDelimiter
+date_YearSuffixImplicitWithNoDateOneOnly
 decorations_SimpleQuotes
 disambiguate_AddNamesSuccess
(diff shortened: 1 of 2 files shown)


PgBiel:
Seems like some of those (such as EnvAndUrb) could warrant an upstream PR to fix the typos, but either way we're kinda "wrong" by reading the `humans` files rather than the `machines` files, which are in JSON and much easier to parse. Regardless, this seems like an improvement, so I'll merge. Thanks!

PR #418, opened 2025-11-07
Allow magic CSL tests

Drodt:
I don't know why these were not allowed before. If there is a good reason, I'm unaware.

Adds 40 tests. 24 pass. No clue how many are skipped.


PgBiel:
Thanks!

PR #421, opened 2025-11-09
Fix author count with et al.

Drodt:
Currently, author count ignore et al. configuration. This PR fixes that.

Adds 3 passing tests.


PgBiel:
Thanks!

PR #426, opened 2025-11-14
Suppress et al in sort

Drodt:
This is not part of the CSL spec but is part of the test suite. It also makes sense.

If you have two items and sort with a name macro with `names-use-first="1"` then you get, e.g.,  "Smith" and "Smith et al." if you render the "et al.". However, then the second is always greater than the first, whereas you want them to be equal and further sort keys to decide which is first.

Makes 1 test pass.


src/csl/rendering/names.rs
```diff
@@ -597,7 +597,7 @@ fn add_names<T: EntryLike>(
                 persons.len() - 1,
             );
         }
-    } else if has_et_al {
+    } else if ctx.instance.sorting.is_none() && has_et_al {
```
PgBiel, line 600: There are other checks for `has_et_al` above. Are we sure we only have to change this step? 

I suppose that there could be a bit of a debate on whether changing the sort behavior too much would be welcome as well. In principle, it seems like the user would find it unexpected for sort to not sort based on the final text, which includes et al. But at least your argument seems reasonable.

Overall, maybe my question is how much do we want to modify the sort to ignore et al.

PgBiel (commented):
Thanks!

src/csl/rendering/names.rs
Drodt: If I change any other check, tests are failing. The behavior seems reasonable to me: Rendering et al for sorting breaks the expected order but all other behavior (et-use-last, etc.) can be controlled by attributes.

Issue #148, opened 2024-04-01
Printing chapters with custom CSL style


ashprice:
Hello, I am trying to format a bibliography in typst. 

Consider the following biblatex reference:
```bib
@incollection{c.r2015,
  title = {The cartography of syntactic structures},
  booktitle = {The {{Oxford}} handbook of linguistic analysis},
  author = {Cinque, Guglielmo and Rizzi, Luigi},
  editor = {Heine, Bernd and Narrog, Heiko},
  date = {2015},
  series = {Oxford handbooks in linguistics},
  edition = {Second edition},
  pages = {51--65},
  publisher = {Oxford University Press},
  location = {New York, NY},
  isbn = {978-0-19-967707-8},
  langid = {english},
  language = {English},
  annotation = {OCLC: ocn908566410}
}
```

And the following CSL:
```xml
<?xml version="1.0" encoding="utf-8"?>
<style class="in-text" version="1.0" and="symbol" et-al-min="5" et-al-use-first="1" initialize="false" initialize-with="" name-as-sort-order="first" demote-non-dropping-particle="never" xmlns="http://purl.org/net/xbiblio/csl">
  <info>
    <title>foobar</title>
    <id>http://www.zotero.org/styles/apa</id>
    <author>
      <name>AMPrice</name>
    </author>
    <updated>2024-03-30T23:11:28+00:00</updated>
    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
  </info>
  <locale xml:lang="en">
    <terms>
      <term name="editor">
        <single>ed.</single>
        <multiple>eds.</multiple>
      </term>
      <term name="translator">
        <single>trans.</single>
        <multiple>trans.</multiple>
      </term>
    </terms>
  </locale>
  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year" givenname-disambiguation-rule="primary-name-with-initials">
    <layout delimiter="; " prefix="(" suffix=")">
      <group>
        <names variable="author" delimiter=", "/>
      </group>
    </layout>
  </citation>
  <bibliography hanging-indent="true" name-as-sort-order="first" et-al-min="21" et-al-use-first="19" entry-spacing="0" line-spacing="1" and="text" initialize-with=" ">
    <layout>
      <group>
        <choose>
[...]

Issue #72, opened 2023-11-01
Error when converting Bibtex to Hayagriva


mihock:
There is a small error in the conversion from Bibtex to Hayagriva. Bibtex items of type `incollection` are converted to Hayagriva items of type `chapter`. However, they should be converted to type `anthos`. On the other hand, Bibtex items of type `inbook` are converted to Hayagriva items of type `anthos`. However, they should be converted to type `chapter`. Both Bibtex items are simply confused in the conversion, so output of the references of these types is wrong in Typst, at least with the APA style. I hope this is enough information to fix the error.

zepinglee:
I seems that both `@inbook` and `@incollection` types from bib(la)tex matches CSL's `chapter` type and they should have no difference in the output.

https://github.com/typst/hayagriva/blob/8dce242de6366156eee2ea113584236883af6e66/src/csl/taxonomy.rs#L565-L567

> Both Bibtex items are simply confused in the conversion, so output of the references of these types is wrong in Typst, at least with the APA style.

Can you provide an example BibTeX entry and the expected output?

mihock:
Here is a small example.

Testfile apatest.typ with two items. The only difference between the two items is that the Bibttex item type is `incollection` (which is correct) or `inbook` which is wrong in this case. With the source of the error, I was not quite correct:  The error seems to occur when the output is converted to pdf, not when the Bibtex file is converted to yml. 

Outputfile: 
[apatest.pdf](https://github.com/typst/hayagriva/files/13235538/apatest.pdf)

The first item (inbook) in the output shows how incollection items should be formatted with apa. The second item (incollection) is formatted in the wrong way.

Testfile:

```
The first item (incollection) should be formatted as an edited book. The correct citation of an edited book is only achieved if the Bibtex item is inbook, which is wrong for this citation.

- @alloy1990_incollection
- @alloy1990_inbook

#bibliography("apatest.yml",
  title: "References",
  style: "apa",
  )
```

Bibtex file:

```
@incollection{alloy1990_incollection,
  title = {incollection: Comorbidity of anxiety and depressive disorders: A helplessness-hopelessness perspective},
  booktitle = {Comorbidity of mood and anxiety disorders},
  author = {Alloy, L.B. and Kelly, K. and Mineka, S. and Clements, C.},
  editor = {Maser, J.D. and Cloninger, C.R.},
  date = {1990},
  pages = {499--543},
  publisher = {{American Psychiatric Press}},
  location = {{Washington, DC}},
[...]

PR #427, opened 2025-11-20
recognize Anthos as Chapter in CSL

Laurendus:
As discussed in #148 a biblatex entry of type `@incollection` is recognized as hayagriva type `Anthos` of an `Anthology`.

#72 (PR #74) claims that `Anthos` is recognized as being a CSL type `chapter`. However https://github.com/typst/hayagriva/issues/148#issuecomment-3064023480 produced a CSL style for debugging this issue which shows, that `@incollection` sources are not recognized as `chapter`s. 

https://github.com/ashprice/hayagriva/commit/49d88b2d37d1e9b72e115bd675ccaaf87bee4101 proposes a solution for this problem, however i could not find a final resolution/verdict or a related PR.

Since I am affected by this problem and it appears to persist I want to propose my solution with a slightly more strict matcher than proposed by https://github.com/ashprice/hayagriva/commit/49d88b2d37d1e9b72e115bd675ccaaf87bee4101.

I hope I did not miss a verdict somewhere else and am open to further opinions on this matter since I am neither an expert in Rust nor in the semantics of citation source types.

### DIFF
--- a/src/csl/taxonomy.rs
+++ b/src/csl/taxonomy.rs
@@ -605,9 +605,10 @@ impl EntryLike for Entry {
 
                 !(is_periodical || is_collection)
             }
-            Kind::Chapter => {
-                select!(Chapter > (Book | Anthology | Proceedings)).matches(self)
-            }
+            Kind::Chapter => select!(
+                (Anthos > Anthology) | (Chapter > (Book | Anthology | Proceedings))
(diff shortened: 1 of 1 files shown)


Drodt (approved):
That makes a lot of sense! Thank you!

Issue #114, opened 2023-11-30
Strange doi presentation with certain bibliography styles


Enivex:
### Description

In springer-mathphys, doi are prefaced by https://doi.org/, but that isn't part of the link, which looks weird.

### Typst version

- [X] I am using the latest version of Typst

yochem:
It seems that Hayagriva just puts in the CSL prefix field for the DOI, which in some cases is `https://doi.org/`, but in others `DOI: `. Thus, hayagriva should make either a distinction between the two (test if the prefix itself is a url) or just include the prefix in the url, also for `DOI: ` prefixes.

TimeTravelPenguin:
I have the same issue. The paper I wanted to cite was `https://dl.acm.org/doi/10.5555/1082161.1082202`. However, using the style _"springer-lecture-notes-in-computer-science"_, the doi `10.5555/1082161.1082202` rendered result was `https://doi.org/10.5555/1082161.1082202`, which is an invalid URL. I have never included the `https://doi.org` prefix, so I have to instead either use `url` (in my `.bib` file) or do something else.

I would really like the ability to remove the prefix. It also makes styling very difficult, as mentioned in a duplicate issue https://github.com/typst/typst/issues/2853

harrysw1729:
> I have the same issue. The paper I wanted to cite was `https://dl.acm.org/doi/10.5555/1082161.1082202`. However, using the style _"springer-lecture-notes-in-computer-science"_, the doi `10.5555/1082161.1082202` rendered result was `https://doi.org/10.5555/1082161.1082202`, which is an invalid URL. I have never included the `https://doi.org` prefix, so I have to instead either use `url` (in my `.bib` file) or do something else.
> 
> I would really like the ability to remove the prefix. It also makes styling very difficult, as mentioned in a duplicate issue [typst/typst#2853](https://github.com/typst/typst/issues/2853)


If you only need to remove the prefix, I think there are three quick workarounds.

The first is to use a different CSL that does not prefix `https://doi.org` to DOIs, e.g., `ieee`.

The second way is to use a nested show-rule, which replaces the `https://doi.org` prefix to `DOI: `.
```typst
#show bibliography: it => {
  show "https://doi.org/": "DOI: "
  it
}
```

[...]

PR #432, opened 2025-12-09
Fix DOI, PMID, and PMCID's not forming proper links with some styles.

peterc-s:
(see #114 for the related issue)

# Problem
Some styles (such as `association-for-computing-machinery`) have `https://doi.org/` as the affix for DOI's. The [CSL 1.0.2 spec](https://docs.citationstyles.org/en/stable/specification.html#appendix-vi-links) says that this should be made part of the link anchor. However, hayagriva doesn't currently do that.

# Solution
I tried to find a good solution to this and document it well. Essentially
1. Check if the CSL style uses the affix given in the standard.
2. If it does, then don't apply the prefix as text. Otherwise, just apply the prefix (as, in the example of `ieee`: this is `doi: ` instead and should not form part of the link)
3. Later, if a `StandardVariable` has a link prefix, check if we skipped applying the prefix. If we did, that means we want the URL prefix displayed on the link. Otherwise, we don't want the URL prefix displayed, but we do want it in the URL itself.

`StandardVariable::URL`s are added without any prefixing and anything else is just pushed (like before).

### DIFF
--- a/tests/citeproc.rs
+++ b/tests/citeproc.rs
@@ -87,6 +87,16 @@ impl fmt::Display for SectionTag {
 enum TestMode {
     Citation,
     Bibliography,
+
+    /// Same as bibliography but made for hayagriva's local tests.
+    /// Includes links, which are normally not printed by citeproc tests.
+    BibliographyFull,
+}
+
+impl TestMode {
+    fn is_bibliography(self) -> bool {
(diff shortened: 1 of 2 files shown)


PgBiel:
Added a local test, to do this I created a new local test mode which also prints rendered links.

Thank you!

Issue #346, opened 2025-07-20
Data fields inconsistencies and outdated test example


jassielof:
# Issues and Inconsistencies with the Documentation and what Typst allows

Tested with:

- Typst v0.13.1.
- Using [Hayagriva's file format](https://github.com/typst/hayagriva/blob/98c3ba49b844921d9004b0a807d25b0114d7df2c/docs/file-format.md) as reference

### `type`

The field is case-insensitive only for the first letter, the rest is just allowed to be lowercase. For example, `WeB` is not allowed, but `Web` is.

### `affiliated`

Typst doesn't throw an error if the `affiliated` field is not an array, but an object.

#### `role`

Typst throws error on any uppercase letter, it only allows lowercase letters.

### `serial-number`

Typst allows these to be number, compared to the docs which only says strings

### `runtime` and `time-range`

These don't seem to be strict or enforce the format.

## Examples

`tests/data/basic.yml` is outdated, line 241 and 47 include translations and sentence-case fields on title.

Issue #36, opened 2023-03-26
Inconsistencies in documentation of volume and issue


larsgw:
The example for journal articles has the `volume` and `issue` fields attached to the direct `parent` of the article, which has the type `Periodical`. I think this might be inconsistent with the rest of the documentation. The description of both the `volume` and `issue` fields state that the field should be attached to the contained entry/item:

> For an item whose parent has multiple issues, indicates the position in the issue sequence. Also used to indicate the episode number for TV.

> For an item whose parent has multiple volumes/parts/seasons ... of which this item is one

This means (if I am interpreting it right) that the fields should be attached to the `Article`, or possibly the entire issue. This means the example could be consistent with the documentation, if the parent entry (where the fields are) represents the entire issue. However, the documentation also seems to state that entries of type `Periodical` represent not just one issue or volume, but the whole journal:

> A publication that periodically publishes issues with unique content. This includes scientific journals and news magazines.

Can a `Periodical` entry also represent individual issues? If not, should the example (or the descriptions) be changed? Or did I simply misunderstand?

PR #438, opened 2025-12-28
Fix file format documentation inconsistency against implementation

jassielof:
Fixes related to #346:
- Makes both the entry and affiliated role types fully case insensitive.
- The file format now explains that the affiliated can be either a single object or a list.
- Removed outdated sentence case element from the basic test.

For the affiliated I initially wanted to not allow single objects, but well, it feels more convenient than having a single element list for it.

For the timestamps, the validation or enforcement regarding the format wasn't valid with the overflow, not sure if this was intended or well 😞.

Also, if might be relevant to fix some examples in #36, with clarification or just commenting on the issue?

### DIFF
--- a/src/types/persons.rs
+++ b/src/types/persons.rs
@@ -35,7 +35,7 @@ impl PersonsWithRoles {
 
 /// Specifies the role a group of persons had in the creation to the
 /// cited item.
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[derive(Clone, Debug, Serialize, PartialEq, Eq, Hash)]
 #[non_exhaustive]
 #[serde(rename_all = "kebab-case")]
 pub enum PersonRole {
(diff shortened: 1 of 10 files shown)


jassielof:
Actually, for the sake of simplicity, I feel that the entire thing of entry and role types shouldn't be much of fully case insensitive now 😔, mainly because of the JSON Schema, at least with the draft 7, it doesn't support case insensitive enumeration, nor case insensitivity in regular expressions. And more up to date drafts/specifications aren't widely or generally supported yet.

For the role types it's easier to make it into the schema with an enumeration, but with the entry types it gets different since one must retype both "article" and "Article" in the enumeration, or well if easier with regular expression just `[Aa]rticle|...` but then it gives a bit of bad feedback for the completions, in editors as it throws that it doesn't match the whole regular expression, compared to a nice list of accepted values. So, for the entry type at least would probably correspond removing the alias? But yeah, _breaking changes_ 😞. 

For the rest, I feel it's fine, maybe? 😭

src/types/persons.rs
```diff
@@ -85,6 +85,58 @@ pub enum PersonRole {
...
+                    "commentator" => Ok(PersonRole::Commentator),
+                    "holder" => Ok(PersonRole::Holder),
+                    "compiler" => Ok(PersonRole::Compiler),
+                    "founder" => Ok(PersonRole::Founder),
+                    "collaborator" => Ok(PersonRole::Collaborator),
+                    "organizer" => Ok(PersonRole::Organizer),
+                    "cast-member" => Ok(PersonRole::CastMember),
+                    "composer" => Ok(PersonRole::Composer),
+                    "producer" => Ok(PersonRole::Producer),
+                    "executive-producer" => Ok(PersonRole::ExecutiveProducer),
```
MultisampledNight, line 125: This doesn't cover `CastMember` or `ExecutiveProducer` as it's still in `file-format.md` (https://github.com/typst/hayagriva/blob/a137441413a5907c15ced44d1502dfb9fa1a3014/docs/file-format.md#affiliated).

MultisampledNight (commented):
Highly appreciated! ^^

src/types/persons.rs
MultisampledNight: I think we could depend on `heck` for the conversion (instead of .to_lowercase, .to_kebab_case) (since it's in our dep tree already it wouldn't impact compile time much)

```
 = cargo tree -i heck
heck v0.5.0
└── strum_macros v0.27.2 (proc-macro)
    └── strum v0.27.2
        └── biblatex v0.11.0
            └── hayagriva v0.9.1 (/home/user/c/d0/software/document/bibliography/hayagriva)
[dev-dependencies]
└── hayagriva v0.9.1 (/home/user/c/d0/software/document/bibliography/hayagriva)
```

jassielof: thanks ill try to implement it once i get a response from the team

Drodt: I agree with @MultisampledNight. Adding `heck` seems like a good idea.

MultisampledNight:
Denoting intent to supersede this PR with a new one the next few days here.

PR #447, opened 2026-01-24
Fix error span pointing to wrong entry when type is missing (#441)

### LINKED ISSUE #441: Wrong confusing span when entry `type` is missing
### Problem

When any given entry in a hayagriva bibliography is missing its `type`, the error span appears to default to the first entry in the entire bibliography, regardless of whether or not it has a `type`.

This makes working with large, possibly faulty, bibliographies *very* frustrating. E.g. I am working with a 2777 line bibliography and recently missed adding a `type: video` to a talk I watched.

### MVCE

Bibliography:

```yaml

a:
  type: article

b:
  title: "Bee movie"
```

Error with `echo '#bibliography("bib.yaml")' | typst c - -`:

```
error: failed to parse YAML (no entry type at line 2 column 1)
  ┌─ bib.yaml:2:0
  │
2 │ a:
  │ ^
```

<details>

<summary>The bare hayagriva error</summary>

```
thread 'main' (361208) panicked at src/main.rs:234:55:
called `Result::unwrap()` on an `Err` value: Error("no entry type", line: 2, column: 1)
```

</details>

AndreyAkinshin:
Fixes #441

When an entry was missing its `type`, the error span would point to the first entry in the bibliography instead of the actual problematic entry. This made debugging large bibliographies difficult.

Root cause: entries were first collected as `NakedEntry` structs, then converted to `Entry` in a second pass. By that point, serde had already moved past the original position.

Solution: use `DeserializeSeed` to convert each entry immediately during deserialization, preserving correct error locations. The error message now also includes the entry key for clarity ("no entry type for `key`").

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -103,4 +103,56 @@ mod tests {
             assert_eq!(match_e, &entry);
         }
     }
+
+    fn assert_error_location(err: &serde_yaml::Error, line: usize, column: usize) {
+        let location = err
+            .location()
+            .unwrap_or_else(|| panic!("Expected error location. Got: {err}"));
+        assert_eq!(
+            (location.line(), location.column()),
+            (line, column),
+            "Error location mismatch. Got: {}:{}",
+            location.line(),
+            location.column()
+        );
+    }
+
+    #[test]
+    fn issue_441() {
+        // Exact scenario from GitHub issue #441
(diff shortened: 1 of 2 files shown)


Drodt (commented):
Looks good to me, thank you! Please merge/rebase and we can get this merged.

PR #450, opened 2026-02-16
[Feature] Pass Directors For Unparented Video and Audio

gozelski:
Currently entries require a video/audio parent to pass the Director affiliated role.
This simple change brings `Director` in line with `SeriesCreator`, allowing unparented video/audio entries to show director attribution on their own.

*Please let me know if this is on purpose / works as intended.*


Drodt (approved):
Makes sense to me. Thank you!

PR #455, opened 2026-03-30
Fix prose citations using "&" instead of "and" between author names

### LINKED ISSUE #454: Prose citations use "&" instead of "and" between authors
Prose citations should use the spelled-out "and" between author names, not "&". This affects styles that specify `and="symbol"` in its CSL such as APA.

- **Parenthetical:** (Wegener & Petty, 1994) (correct)
- **Narrative/prose:** Wegener and Petty (1994) (incorrect, should use "and", currently uses "&")

For APA specifically, this is per the APA manual (7th ed., Section 8.17) and Purdue OWL: https://owl.purdue.edu/owl/research_and_citation/apa_style/apa_formatting_and_style_guide/in_text_citations_author_authors.html

This is also a follow-up to #243, where the original reporter expected "Brady and Collier (2010)" and the fix there resolved the missing editor names but the test asserted "&" instead of "and".

### Reproduction

`test.bib`:
```bibtex
@article{wegener94,
  author = {Wegener, Duane T. and Petty, Richard E.},
  title = {Mood Management Across Affective States},
  journal = {Journal of Personality and Social Psychology},
  year = {1994},
  volume = {66},
  pages = {1034--1048},
}
```

`test.typ`:
```typst
Prose citation: #cite(<wegener94>, form: "prose")

Parenthetical citation: #cite(<wegener94>)

#bibliography("test.bib", style: "apa")
```

### Actual output

> Prose citation: Wegener & Petty (1994)
>
> Parenthetical citation: (Wegener & Petty, 1994)

### Expected output

> Prose citation: Wegener and Petty (1994)
>
> Parenthetical citation: (Wegener & Petty, 1994)

### Version

typst 0.14.2

minco3:
- Prose citations (form: "prose") currently use "&" to join author names, but should use "and". This affects any CSL style with and="symbol" (e.g. APA).
  - Adds a use_text_and flag to WritingContext, set during prose author rendering, which causes the name joiner to use the localized "and" term instead of "&"
  - Updates the existing issue_243 test to expect "and"

The approach works and matches the existing pattern of toggling context flags in WritingContext, though it's admittedly a bit of a bolted-on solution. Open to suggestions on a cleaner approach.

Closes #454

### DIFF
--- a/src/csl/mod.rs
+++ b/src/csl/mod.rs
@@ -1377,7 +1377,9 @@ impl<'a> StyleContext<'a> {
                 }
             }
             (Some(CitePurpose::Prose), _) => {
+                ctx.writing.use_text_and = true;
                 do_author(&mut ctx);
+                ctx.writing.use_text_and = false;
                 if !self
                     .csl
                     .citation
@@ -1735,6 +1737,8 @@ pub(crate) struct WritingContext {
     /// Delimiters from ancestor delimiting elements (e.g., `cs:group`).
     /// To be applied within `cs:choose`, but not another delimiting element or a macro.
     delimiters: NonEmptyStack<Option<String>>,
+    /// Use "and" instead of "&" when joining names, for prose citations.
+    use_text_and: bool,
 
     // Buffers.
     /// The buffer we're writing to. If block-level or formatting changes, we
@@ -1760,6 +1764,7 @@ impl Default for WritingContext {
             cases: NonEmptyStack::default(),
             name_options: NonEmptyStack::default(),
             delimiters: NonEmptyStack::default(),
+            use_text_and: false,
             buf: CaseFolder::default(),
             elem_stack: NonEmptyStack::default(),
         }
@@ -3347,7 +3352,7 @@ mod tests {
             .unwrap();
 
         assert_eq!(c1, "Downs (1957)");
-        assert_eq!(c2, "Brady & Collier (2010)");
+        assert_eq!(c2, "Brady and Collier (2010)");
     }
     #[test]
     #[cfg(feature = "archive")]
--- a/src/csl/rendering/names.rs
+++ b/src/csl/rendering/names.rs
@@ -542,6 +542,10 @@ fn add_names<T: EntryLike>(
                         NameAnd::Text => ctx
                             .term(Term::Other(OtherTerm::And), TermForm::default(), false)
                             .unwrap_or_default(),
+                        // Prose citations use "and" instead of "&".
+                        NameAnd::Symbol if ctx.writing.use_text_and => ctx
+                            .term(Term::Other(OtherTerm::And), TermForm::default(), false)
+                            .unwrap_or("and"),
                         NameAnd::Symbol => "&",
                     });
                     ctx.ensure_space();
@@ -552,6 +556,9 @@ fn add_names<T: EntryLike>(
                         NameAnd::Text => ctx
                             .term(Term::Other(OtherTerm::And), TermForm::default(), false)
                             .unwrap_or_default(),
+                        NameAnd::Symbol if ctx.writing.use_text_and => ctx
+                            .term(Term::Other(OtherTerm::And), TermForm::default(), false)
+                            .unwrap_or("and"),
                         NameAnd::Symbol => "&",
                     });
                     ctx.ensure_space();


Drodt:
Thank you for this PR!

Before merging, I want to make sure that this does not break other styles. I could not find anything in the CSL spec about this. Is this just a quirk of the APA style? Or do you have any information that this is a general solution?

minco3:
Thanks for the reply! Looking back after your comment, the current PR is too broad. It affects any CSL style with `and="symbol"`, not just APA.

A few findings from digging into the CSL spec and citeproc-js:

- CSL 1.0.2 has no concept of prose/narrative citations. The and attribute applies universally per `<cs:name>`, so this isn't something a style author can express in standard CSL today. https://discourse.citationstyles.org/t/different-and-for-narrative-and-parenthetical-citations/1453/5
- citeproc-js handles this via `cs:intext` (a sibling to `cs:citation` that defines separate name rendering for prose forms. citeproc-js's own docs call it "outside the official CSL Specification"). It's on the [CSL 1.1 branch](https://github.com/citation-style-language/schema/issues/365) but 1.1 hasn't shipped, and hayagriva doesn't currently support it either way.
- Within hayagriva, the existing prose-mode code already branches on `StyleCategory::CitationFormat` (Label/Numeric vs author-date), so there's precedent for narrowing behavior by style category.

Full disclosure: I'm not a researcher and the bulk of my APA experience is from school assignments, so I'd value your call on how you want to handle this. Some options I see:

1. Implement `cs:intext` for proper per-style control. Cleanest, but a much larger change touching citationberg. (similar to what citeproc-js does)
2. Narrow the current PR to only flip the joiner for author/author-date styles (using the existing StyleCategory pattern), since `and="symbol"` mostly only manifests in those formats.
3. Expose it as an opt-in flag at the citation request level, pushing the decision to typst.

Drodt:
Thanks for looking into it! 

At the moment, I would like to hold off on merging this. We will need it at some point for sure, but I am wary of diverging too much from the CSL spec. If CSL 1.1's solution changes before it is stable, that will just introduce way too many headaches. And since styles cannot really support it yet anyway, I see no rush.

So I suggest we wait until CSL 1.1 is ready. Hopefully, that's not a big problem for your use case.

mm-tea:
Are you sure this is the intended behavior in all styles that use `and="symbol"`? I have come across [this style specification](https://www.eva.mpg.de/lingua/pdf/GenericStyleRules.pdf) which explicitly calls for "&" in prose-style citations (see pages 5-6). I agree with this change being too broad -- I'd be careful of assuming APA is representative of all (or even most) styles out there.

minco3:
Agreed, this is too broad. @mm-tea's example of an author-date style that deliberately wants "&" in prose shows there's no blanket rule that's safe. The choice is per-style, and the only place to encode it in the XML is something like cs:intext.

The catch with waiting is CSL 1.1 has no release timeline, and cs:intext has sat unchanged on the schema branch since 2020. So "wait for 1.1" could mean indefinitely, with no correct narrative output for APA in the meantime.

As a middle ground, would you be open to an opt-in instead of a default change? Something like a flag on the citation item, off by default, that tells the renderer to spell out "and" in prose citations. Nothing changes for any existing style, the caller asserts the per-style intent instead of hayagriva guessing, and APA narrative renders correctly.

It would sit alongside the existing per-item controls like hidden and purpose, and since CitePurpose::Prose is already a hayagriva extension with no CSL equivalent. This is just another knob on a feature that already lives outside the spec. If cs:intext gets merged to main, it supersedes the flag cleanly. I'd be glad to rework the PR along those lines.

Drodt:
I don't really see a good solution for this at the moment. So long as the APA CSL style does not consider `cs:intext`, there is nothing we can really do. Setting a flag on a per citation basis would fine in principle but how is this value set? We would need to hard code it for APA and that does not seem like a very solid solution.

But I'm open to change my mind. Do you see any other solution?

minco3:
Fair, the flag has exactly that problem. Someone still has to know APA wants it, which means hardcoding style IDs somewhere.

What about option 1 from my earlier comment: supporting `cs:intext` itself? It feels like the right approach because the value lives in the style's XML instead of being hardcoded anywhere. citationberg would parse the `<intext>` element and hayagriva would honor that layout when a style provides one, falling back to today's `&` output when it doesn't. Nothing becomes APA-specific as it's just another layout alongside citation and bibliography, and it matches both what the CSL 1.1 draft defines and what citeproc-js already implements.

It won't change stock APA on its own, since the published APA.csl has no intext block yet. But that's the right outcome as the fix belongs in the style data, and this keeps hayagriva from being the blocker. The moment APA or a user's variant adds the block, it works with no changes here.

If you're on board I'm happy to take a stab at it, probably starting with a citationberg PR since hayagriva depends on it.

Drodt:
Feel free to implement it. But I recommend waiting until #489 is merged, which should be soon-ish. Merge conflicts with that are a pain.

PR #456, opened 2026-04-04
Fix name suffix example in documentation

### LINKED ISSUE #451: Name suffix not handled properly
A person's name suffix is not handled properly if the name is specified as a single string.

As per [the documentation](https://github.com/typst/hayagriva/blob/main/docs/file-format.md#Person), the format of a name with a suffix should be written as `last name, given name, suffix`. However, this results in incorrect formatting in multiple bibliography styles, unlike explicitly marking the suffix with subfields:

The bibliography file
```yaml
stewart1:
  type: Thesis
  author: Stewart, Thomas W., Jr.
  date: 2000
  title: "Mutation as morphology: Bases, stems, and shapes in Scottish Gaelic"
  place: Columbus, OH
  publisher: The Ohio State University
  genre: dissertation

stewart2:
  type: Thesis
  author:
    name: Stewart
    given-name: Thomas W.
    suffix: Jr.
  date: 2000
  title: "Mutation as morphology: Bases, stems, and shapes in Scottish Gaelic"
  place: Columbus, OH
  publisher: The Ohio State University
  genre: dissertation
```
produces two different bibliography items, with only the latter being correct (in this case, using `"mla"` style, but the same occurs with other citation styles, e.g. `"apa"` or `"ieee"`):

Image

rather, it seems to me that formatting the author name as `last name, suffix, given name` produces the correct name formatting in most styles, so this might just be a documentation issue.

mm-tea:
[The documentation for the person data type](https://github.com/typst/hayagriva/blob/main/docs/file-format.md#Person) shows an incorrect example of a name with a suffix - `Luther King, Martin, Jr.` instead of the correct `Luther King, Jr., Martin`.

This pull request changes this example to the correct name order.

Closes #451

### DIFF
--- a/docs/file-format.md
+++ b/docs/file-format.md
@@ -452,7 +452,7 @@ journal:
 
 #### Person
 
-A person consists of a name and optionally, a given name, a prefix, and a suffix for the (family) name as well as an alias. Usually, you specify a person as a string with the prefix and the last name first, then a comma, followed by a given name, another comma, and then finally the suffix. Following items are valid persons:
+A person consists of a name and optionally, a given name, a prefix, and a suffix for the (family) name as well as an alias. Usually, you specify a person as a string with the prefix and the last name first, then a comma, optionally followed by a suffix, another comma, and then finally the given name. Following items are valid persons:
 
 - `Doe, Janet`
 - `Luther King, Jr., Martin`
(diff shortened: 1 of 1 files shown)


Drodt:
Thanks!

PR #475, opened 2026-05-26
Pin actions

laurmaedje:
To increase supply chain security.

Pinning done via https://github.com/mheap/pin-github-action.


Drodt:
It seems the pinned Rust version is no longer available? Tests are failing

laurmaedje:
GitHub actions was down earlier today.

PR #479, opened 2026-05-27
Add citeproc tests that failed because of `<b> </b>`

YDX-2147483647:
These three citeproc tests actually pass, but there were thought as failed because hayagriva writes ` ` (whitespace) as `<b> </b>`.
This PR adds them to `citeproc-pass.txt`.

[`name_InTextMarkupInitialize`](https://github.com/citation-style-language/test-suite/tree/master/processor-tests/humans/name_InTextMarkupInitialize.txt) and [`name_InTextMarkupNormalizeInitials`](https://github.com/citation-style-language/test-suite/tree/master/processor-tests/humans/name_InTextMarkupNormalizeInitials.txt) have a similar problem.
They provide the author name as `{ "given": "<b>John</b>-Quiggly" }`, and hayagriva generates `<b>J. Q.` instead of the expected `<b>J.</b> Q.`.
However, that's more difficult to circumvent, and they are the only two test cases whose names contain _InTextMarkUp_.
Therefore, I don't they worth special handling.

(By the way, I found them with https://ydx-2147483647.github.io/haya-view-test/.)

### DIFF
--- a/tests/citeproc-pass.txt
+++ b/tests/citeproc-pass.txt
@@ -487,6 +487,9 @@ sort_NumberOfAuthorsAsKey
 sort_StatusFieldAscending
 sort_StatusFieldDescending
 sort_TestInheritance
+sort_VariousNameMacros1
+sort_VariousNameMacros2
+sort_VariousNameMacros3
 sortseparator_SortSeparatorEmpty
 substitute_RepeatedNamesOk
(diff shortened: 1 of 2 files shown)


Drodt (approved):
Thanks! And that website is amazing! I was considering setting something like that up.

PR #487, opened 2026-06-12
Fix citation number offsetting for sorted bibliographies

laurmaedje:
https://github.com/typst/hayagriva/pull/409 only handled citation number offsetting for styles without sorting. This PR extends it to sorted styles.


Drodt (approved):
Looks good.

PR #495, opened 2026-07-09
Describe the data types missing in docs/file-format

### LINKED ISSUE #493: YAML format: missing docs for "url" and "publisher" type
The [Fields](https://github.com/typst/hayagriva/blob/main/docs/file-format.md#fields) section of the file format reference refers to types "url" and "publisher".

The "url" type is implied by the example and its usage elsewhere in the document to be either a string or a map like `{ value: String, date: Date }`.

The "publisher" type is implied by the example and its usage elsewhere in the document to be either a string or a map like `{ name: String, location: String}`, but these could possibly be FormattableStrings instead?

Both types should have an entry in the [Data types](https://github.com/typst/hayagriva/blob/main/docs/file-format.md#data-types) section.

YDX-2147483647:
Resolves #493.

Publisher and URL are mentioned in the _Fields_ section, but are missing in the _Data types_ section.

Reference: [`hayagriva::types` - Rust](https://docs.rs/hayagriva/0.10.1/hayagriva/types/index.html)

---

_London_ in the example for `publisher.location` is changed to _London, UK_, because the latter format is more consistent with test cases across the current code base.

https://github.com/typst/hayagriva/blob/2478d0741618d4c2ba34ba3080def9ca5f03d1f2/tests/data/basic.yml#L201
https://github.com/typst/hayagriva/blob/2478d0741618d4c2ba34ba3080def9ca5f03d1f2/tests/data/basic.yml#L218
https://github.com/typst/hayagriva/blob/2478d0741618d4c2ba34ba3080def9ca5f03d1f2/src/io.rs#L22


clbarnes (approved):

Drodt (approved):
Thanks!

PR #496, opened 2026-07-09
Add ability to declare locator type to CLI

### LINKED ISSUE #492: CLI: Add ability to declare locator type
Currently all locators are being treated as `Locator::Custom` which causes some citations to not be formatted correctly. For example, MLA citations will be formatted (author, page number), which is incorrect by MLA standards because of this (there shouldn't be a comma). This was also mentioned by issue typst/typst#8609. Allowing the user to declare the type of their locator would fix this issue.

landenhord:
This pull request adds the ability for the user to specify locator type in the CLI through the use of a `--type` flag. Before, the CLI treated all locators as `Locator::Custom` which caused improper formatting for some citation styles, such as MLA. Resolves #492

Change in use:
<pre>
hayagriva cite example.yaml --style mla -k example --locators 8 <strong>--type page</strong>
</pre>
Which would result in `(Example 8)` in contrast to the previous `(Example, 8)`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -127,6 +127,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1ddb117e43bbf7dacf0a4190fef4d345b9bad68dfc649cb349e7d17d28428e51"
 dependencies = [
  "clap_builder",
+ "clap_derive",
 ]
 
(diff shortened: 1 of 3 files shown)


Drodt (approved):
Thanks!

PR #502, opened 2026-07-24
chore: replace deprecated serde-yaml with yaml_serde

BrainTeazer:
serde-yaml is no longer maintained: https://github.com/dtolnay/serde-yaml
The YAML organization actively maintains a fork called yaml_serde: https://github.com/yaml/yaml-serde

There are two recommended options for migration, use Cargo's package renaming and update all imports directly (done here).


Drodt:
@laurmaedje You assigned this to yourself, so just to give my two cents: Merging this seems like a good idea. The fact that the YAML organization maintains it, seems like a good sign to me.

Drodt (approved):
I think it makes sense. I see some active development, but nothing fishy, and it is by the official YAML org.

Thank you!

Issue #246, opened 2024-10-29
APA date not working properly with bibtex and YAML


gkourgkoutas:
### Description
When I'm using my bibliography and I choose APA7 as cite style, it doesn't matter if I use a biblatex file or YAML (typst hayagriva). In both cases, if the date (year) is added, there are some formatting issues.

Bibtex file:
```latex
@online{nistCVE,
	author = {NIST},
	title = {CVE-2021-44228},
	url = {https://nvd.nist.gov/vuln/detail/CVE-2021-44228},
        date = {2021},
	urldate = {2024-10-18},
}
```

YAML:
```yaml
nistCVE:
 author: "NIST"
 type: "Web"
 title: "CVE-2021-44228"
 date: "2021"
 url: 
    value: "https://nvd.nist.gov/vuln/detail/CVE-2021-44228"
    date: 2024-10-28
```

These produce the following output:

<img width="885" alt="grafik" src="https://github.com/user-attachments/assets/d02dfb2e-f0a1-4139-b79a-3cc7c3752663">

While the output seems fine, there is no "Retrieved + Date", which is the case in latex when `\urldate` is used.

Note that there is also a comma after the year (2021,) which should not be there if only the year of publication is given. typst expects a full date like 2021-01-01.

If the date (publication year or year in bibtex) is removed...

```latex
@online{nistCVE,
	author = {NIST},
	title = {CVE-2021-44228},
	url = {https://nvd.nist.gov/vuln/detail/CVE-2021-44228},
	urldate = {2024-10-18},
}
```
```yaml
nistCVE:
 author: "NIST"
 type: "Web"
 title: "CVE-2021-44228"
 url: 
    value: "https://nvd.nist.gov/vuln/detail/CVE-2021-44228"
    date: 2024-10-28
```

... it shows **"Retrieved + Date"** what I want to achieve.

<img width="903" alt="grafik" src="https://github.com/user-attachments/assets/e44035c9-e8be-457e-b491-fbe678c39953">

**However**, as you can see in the screenshot above, after the author, **the year is missing!**.
The desired output should look like this:

<img width="834" alt="grafik" src="https://github.com/user-attachments/assets/0265a115-0801-4601-9dc7-04b9e9360726">

### Steps to reproduce
- Create a project in typst, the output can be replicated in the **web application as well as locally** (I use mac + homebrew).
- Create a .bib or .yml file for the bibliography.
- Use the date (year of publication) for the reference + URL DATE (the date the site was accessed).

[...]

PR #503, opened 2026-07-31
Fix trailing comma on APA online citation

jassielof:
Ok, first of all, sorry for the previous summary, it was generated with Copilot and was for my own use...

The changes _should_, basically, fix the respective issue mentioned in the test (#246), the trailing comma on online dates with only a year in APA style.


Drodt (approved):
That makes a lot of sense. Thanks!

PR #508, opened 2026-08-09
Always keep serial numbers as verbatim

### LINKED ISSUE #506: serial-number containing hyphens is parsed as a number range
`se.yaml`

```yaml
rep:
  type: report
  title: A numbered release
  author: Hana Hill
  serial-number: USDL-26-0599
  date: 2024
```

`se.typ`

```typst
#bibliography("se.yaml", style: "ieee", full: true)
```

```
actual:   [1] Hana Hill, "A numbered release," USDL-26–599, 2024.
expected: [1] Hana Hill, "A numbered release," USDL-26-0599, 2024.
```

Image

Two corruptions: the second hyphen becomes an en dash, and the leading zero of `0599` is dropped. The value looks like it is taken for a numeric range, `USDL-26` to `0599`. `USDL-26-0599` is a real identifier, a US Bureau of Labor Statistics news release.

Same from a `.bib` source with `number = {USDL-26-0599}`, and with `apa`, `cse` and `nlm-citation-sequence`. Braces (`{{USDL-26-0599}}`) and a non-breaking hyphen make no difference.

Both reference implementations of the `.bib` format keep the value verbatim from an identical `number = {USDL-26-0599}`: bibtex with `plain.bst` gives "Technical Report USDL-26-0599", biblatex with biber gives "Tech. rep. USDL-26-0599".

Two related issues, neither of which covers this:

* #440 was closed as completed on 2026-05-26 for exactly the leading-zero loss, with the `i32` storage named as the cause. v0.10.0 shipped on 2026-06-05 and v0.10.1 on 2026-06-14, so that fix is in the version tested here, yet the zero is still dropped. The range path may bypass it.
* #439 argues from CSL 1.0.2 that numeric detection is too permissive. Under that reading `USDL-26-0599` should not be treated as numeric at all, which would remove both corruptions at once.

Tested against the released 0.10.1 as bundled in Typst 0.15.1 (9dfd3a08). I have no Rust toolchain here, so I could not check `main`.

YDX-2147483647:
Resolves #506

In the current `docs/file-format.md`, the type of `serial-number` is specified as _string_, not _numeric or string_:
https://github.com/typst/hayagriva/blob/0eaceabb18ae034eb8bc284d541810d69ba5f2a6/docs/file-format.md?plain=1#L331

However at present, hayagriva tries to parse serial numbers as numeric numbers when .bib → .yaml and when .yaml → CSL variable.
Usually it fails, so serial numbers will be kept as verbatim as expected; but when it succeeded, the serial number is transformed unexpectedly.

#506 gives an example of the latter case: USDL-26-0599.
Without knowing it's a serial number, we may interpret it as a range from number 26 (with prefix USDL-) to number 599, and format it incorrectly as USDL-26–599 (USDL-26, en dash, 599).

To fix it, this PR removes the parsing logic for serial numbers from `interop.rs` (.bib → .yaml) and `taxonomy.rs` (.yaml → CSL variable).

### Additional info

Leading zeros in serial numbers should be kept, but I'm not sure whether leading zeros in volumes should be kept.
Therefore, I intentionally construct test cases that do not involve volumes with leading zeros in this PR.
I or someone else may deal with it in future PRs.


Drodt (approved):
Thank you!

PR #512, opened 2026-08-25
Avoid empty parentheses in prose citations

### LINKED ISSUE #347: Citation in MLA includes empty parentheses
# Description
When using a block quote with MLA-style citation, the citation will produce an empty set of parentheses.

# Steps to Reproduce

main.typ
```typst
#quote(attribution: <pratchett96>, block: true)[
  No enemies had ever taken Ankh-Morpork. Well, technically they had, quite
  often; the city welcomed free-spending barbarian invaders, but somehow the
  puzzled raiders always found, after a few days, that they didn't own their
  horses anymore, and within a couple of months they were just another minority
  group with its own graffiti and food shops.
]

#bibliography("main.bib", style: "mla")

```

main.bib
```bib
@book{pratchett96,
    title={Eric},
    author={Pratchett, T.},
    isbn={9780575600010},
    series={Discworld series},
    url={https://books.google.com/books?id=UIdkPwAACAAJ},
    year={1996},
    publisher={Vista}
}
```

This produces the following output:
Image

Ideally, the citation should not include the set of empty parentheses, and simply have the author last name without the year entirely, which would align with how MLA citation works outside of block quotes.

SathyaTadinada:
Fix prose citations that produce empty parentheses when suppressing the author leaves no remaining content.

Using the example from Issue #347, an MLA prose citation now renders `Pratchett` instead of `Pratchett ()`.

The prose renderer now tracks whether the author was rendered and conditionally retains the parenthetical portion only when the suppressed-author citation contains content. Existing numeric, label, and authorless citation behavior is preserved.

- Added a regression test for MLA prose citations (`fn issue_347() {...}`)
- Ran all library tests
- Ran `cargo clippy --lib --features csl-json`
- Ran `cargo fmt --check --all`

Closes #347.


Drodt (approved):
Thank you!

PR #515, opened 2026-08-29
Replace Harry Potter with Inheritance Cycle

jorins:
This PR replaces references to Harry Potter with ones to the Inheritance Cycle.

Many of my friends are fleeing Britain and the United States in the wake of transphobic persecution. It is well known that Rowling finances a lot of transphobia. Seeing Harry Potter referenced as the very first example when clicking the docs link from [typst](https://typst.app/docs/reference/model/bibliography/) is a minor gut punch.

I picked the Inheritance Cycle because it was the first example I could think of that has a straightforward number of volumes and distinct chapters, making for a 1:1 replacement of the existing data structure without supplying inaccurate or incomplete data.


Drodt:
I like this change. But I would like the title of the book to be pronounceable. Could you change it to the fourth volume, _Inheritance_?

jorins:
> I like this change. But I would like the title of the book to be pronounceable. Could you change it to the fourth volume, _Inheritance_?

Changed! I was unable to set correct the page interval, but I'm guessing that's not particularly relevant to the test suite :)

docs/file-format.md
```diff
@@ -8,13 +8,13 @@ In technical terms, a Hayagriva file is a YAML document that contains a single m
...
 ```yaml
-harry:
+inheritance:
     type: Book
-    title: Harry Potter and the Order of the Phoenix
-    author: Rowling, J. K.
-    volume: 5
-    page-total: 768
-    date: 2003-06-21
+    title: Inheritance
```
YDX-2147483647, line 13: Perhaps we should pick another example that shows the `title` field should be in title case, not sentence case? _Inheritance_ is a single word. It cannot distinguish them.

**Update:**
Oh, there're two examples here. The other example does imply the format of `title`.

**Update 2:**
I've just checked [Zotero Documentation](https://www.zotero.org/support/kb/sentence_casing) and [CSL 1.0.2 Specification](https://docs.citationstyles.org/en/stable/specification.html#sentence-case-conversion). It's recommended to store the title as sentence case in CSL:

> Unfortunately, it’s not possible for Zotero or any other tool to automate conversion to sentence case in a way that’s 100% reliable, since a computer can’t know for sure that something is a proper noun. (Think, for example, of the word “united”, which can be an adjective, verb, name of a company, or part of a country name.)

> CSL processors don’t recognize proper nouns. As a result, strings in sentence case can be accurately converted to title case, but not vice versa. For this reason, it is generally preferable to store strings such as titles in sentence case, and only use `text-case` if a style desires another case.

Does it apply to Hayagriva as well? If so, we should edit the other example in `docs/file-format.md`.

jorins: The book does have a subtitle, "The Vault of Souls":
image

I'm not sure how best to add/format that in the markup or if it should be added at all, but it would be sufficient to showcase title case.

Drodt: I think the other example is sufficient. And as far as I can tell, Hayagriva handles sentence case decently as well.

Adding a subtitle starting with "or" is a bit annoying in citation formats. Just leave it as is.

Drodt (approved):
Thanks!

jorins:
Awesome! Cheers :)

PR #518, opened 2026-08-29
Bump Rust to 1.98.0 on CI

Drodt:
Follows Typst's current version


laurmaedje:
We should also add a workflow to check the Minimum Supported Rust version here because Typst's is still [1.92](https://github.com/typst/typst/blob/6a6f1a97a3446fc6ba5a723b0baf20a0cfeaaef5/Cargo.toml#L8) and it would be unfortunate if usage of some feature beyond Typst's MSRV would slip in.

There will likely also be a 1.98.1 which we should upgrade to since 1.98.0 has a critical miscompilation bug (though it's unlikely to trigger here). I wasn't aware of that when upgrading the version in Typst.

PR #521, opened 2026-08-31
Add workflow to check the Minimum Supported Rust version

Drodt:
As mentioned here: https://github.com/typst/hayagriva/pull/518#issuecomment-5474782405, Typst's MSRV is 1.92, so Hayagriva should also support that version. This should be checked in the CI.


.github/workflows/ci.yml
```diff
@@ -29,3 +29,12 @@ jobs:
...
       - run: cargo doc --workspace --no-deps
+
+  min-version:
+    name: Check minimum Rust version
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      - uses: dtolnay/rust-toolchain@87eb139fed4b08a67bd1fa429a21d1f5d523e03e # 1.92.0
+      - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
+      - run: cargo check --workspace
```
laurmaedje, line 40: Maybe we should also add `--all-features` here

laurmaedje: (Would also be a good call for Typst itself where I suppose this comes from :D)

Drodt: Yep, it's from Typst. But we should definitely add it. See #522.

PR #523, opened 2026-09-05
Fix delimiter between collapsed and grouped cites

### LINKED ISSUE #379: Fix delimiter between collapsed and grouped cites
### Description

In many author-date citation styles (e.g. APA, APSA, ASA), cites by the same author are collapsed into `(Jones 1999; Smith 2000, 2001)`. But typst renders this as `(Jones 1999; Smith 2000; 2001)`.

After some digging, I found that this is partly caused by the messy schema of CSL (see: https://github.com/citation-style-language/test-suite/issues/36#issuecomment-667126525). Since "Delimiter between collapsed cites" is "Not explicit; will pick up group delimiter if citations are sorted", it seems that typst is incorrectly picking up `delimiter` on `cs:citation`, which is `; ` instead of `, ` in the above mentioned styles.

I tried to work around this by adding a `cite-group-delimiter=", "` to the CSL file. This produces incorrect results like `(Doe 1998, Jones 1999; Smith 2000, 2001; Wilson 2002)`, and `(Doe 1990, Jones 1999, Smith 2000)`. So it seems typst is not [grouping cites](https://docs.citationstyles.org/en/stable/specification.html#cite-grouping) correctly either.

Before this is fixed, a large number of citation styles are just unusable in typst.

### Reproduction URL

_No response_

### Operating system

Web app

### Typst version

- [x] I am using the latest version of Typst

fredguth:
Fixes #379.

Author-date styles that collapse cites by the same author rendered `(Jones 1999; Smith 2000; 2001)` instead of `(Jones 1999; Smith 2000, 2001)`. Two independent problems were involved.

### 1. Default value of `cite-group-delimiter`

`cite-group-delimiter` was only honoured when the style set it explicitly; otherwise the layout delimiter was used *within* a cite group. The CSL spec gives the attribute a default of `", "`.

That default cannot be applied unconditionally, though: per [citation-style-language/test-suite#36](https://github.com/citation-style-language/test-suite/issues/36) it only takes effect for styles that sort their citations. Test cases such as `disambiguate_YearCollapseWithInstitution`, `disambiguate_YearSuffixWithEtAlSubsequent` and `date_YearSuffixDelimiter` have no `cs:sort` on `cs:citation` and do expect the layout delimiter — applying `", "` everywhere breaks all three. The default is therefore applied only when `cs:citation` carries a `cs:sort`, which is the case for APA, ASA and APSA.

### 2. Group delimiter leaking in front of single-cite groups

Cite grouping wrote the group delimiter onto the *first* cite of every group. Since `delim_override` is the delimiter rendered *before* an item, this leaked the group delimiter in front of groups with a single cite: a style that sets `cite-group-delimiter=", "` explicitly produced `(Doe 1998, Jones 1999; Smith 2000, 2001; Wilson 2002)`, as reported in the issue.

The same loop had three further problems, fixed along the way:

- it cleared the delimiter of the previous group member, so groups of three or more lost the delimiter on their middle cites;
- group numbers came from a global counter incremented on first sight of a name, so a cite joining an earlier group was tagged with the wrong group;
- the index recorded for a group went stale once a later cite was moved past it.

The group map now tracks `(index of last group member, group number)` per names elem and fixes up the indices shifted by a move.

### Testing

`bugreports_SortSecondaryKey` now passes and is added to `tests/citeproc-pass.txt`; no test regressed in the citeproc suite. Two local tests, `collapse_YearGroupDelimiterDefault` and `collapse_YearGroupDelimiterExplicit`, cover the implicit and the explicit `cite-group-delimiter` case respectively — both fail without this change.


laurmaedje:
Thanks for the PR. Unfortunately, we do not accept PRs that are implemented or described with AI. This is currently unfortunately only spelled out in our main repository's [`CONTRIBUTING.md`](https://github.com/typst/typst/blob/main/CONTRIBUTING.md).

PR #524, opened 2026-09-06
Respect the `plural` attribute on page labels

### LINKED ISSUE #377: CSL `plural="never"` is ignored
See https://github.com/typst/hayagriva/issues/376 for some context and example sources.

```sh
hayagriva items.yaml reference -n --style gost-r-705-2008-numeric --locale en-US
```

This creates

> Mares I. Firms and the welfare state: When, why, and how does social policy matter to employers? // Varieties of capitalism. The institutional foundations of comparative advantage / ed. Hall P.A., Soskice D. New York: Oxford University Press, 2001. Pp. 184–213.

where Pp is not how `plural="never"` works, nor how it is formatted in https://editor.citationstyles.org/visualEditor/. It must format it as "P" for en-US, and "С" for ru-RU.

https://github.com/citation-style-language/styles/blob/371ebd448591d48689600b110605c884fedad7fe/gost-r-7-0-5-2008-numeric.csl?plain=1#L145

I did find https://github.com/typst/hayagriva/blob/ea4c06a04742f98b92d70ceaedbae7c2d5d017d1/CHANGELOG.md?plain=1#L39-L40 (https://github.com/typst/hayagriva/pull/145), but now I don't think it broke it.

Was found together with: https://github.com/typst/hayagriva/issues/376

MaxFreedomPollard:
Resolves #377.

A `cs:label` that points at a page variable ignores the `plural` attribute, so `plural="never"` still renders "pp." and `plural="always"` still renders "p.". With `gost-r-705-2008-numeric`, which sets `plural="never"` on its page label, a chapter on pages 184 to 213 comes out as "Pp. 184–213" instead of "P. 184–213".

The cause is in `impl RenderCsl for citationberg::Label` in `src/csl/rendering/mod.rs`. The `NumberOrPageVariable::Number` branch runs the label through `label_pluralization`, which reads `label.label.plural`. The `NumberOrPageVariable::Page` branch does not: line 524 in `render` and line 586 in `will_have_info` both compute `let plural = p.as_typed().is_some_and(|p| p.is_plural());`, which looks only at the page value and never at the attribute.

This adds `page_label_pluralization` next to `label_pluralization`, applying the same always/never/contextual rules to page variables, and calls it from both places.

Verification, on macOS with cargo 1.95:

`cargo run --features cli,csl-json -- items.yaml reference -n --style gost-r-705-2008-numeric --locale en-US` on the entry from #377 now ends with "New York: Oxford University Press, 2001. P. 184–213." It printed "Pp. 184–213." before.

The new fixture `tests/local/label_Issue377.txt` renders three page labels (`plural="never"`, `plural="always"`, and the contextual default) over a range and over a single page. On an unmodified `main` it fails, producing "pp. pp. pp. 184–213" and "p. p. p. 184" against the expected "p. pp. pp. 184–213" and "p. pp. p. 184". With the fix, `cargo test --features csl-json --test citeproc test_local_files` passes.

`cargo test --features csl-json` passes in full: 62 unit tests, 4 doc tests, 2 archiver tests, and 6 citeproc tests. That includes `test_parse_tests`, which runs the upstream citeproc test suite and compares it against `tests/citeproc-pass.txt`, so no test in that list changed state and the file needs no update. `cargo fmt --check --all`, `cargo clippy --workspace --all-targets --all-features` and `cargo doc --workspace --no-deps` are clean with `RUSTFLAGS=-Dwarnings` and `RUSTDOCFLAGS=-Dwarnings`.


laurmaedje:
Thanks for the PR. Unfortunately, we do not accept PRs that are implemented or described with AI. This is currently unfortunately only spelled out in our main repository's [`CONTRIBUTING.md`](https://github.com/typst/typst/blob/main/CONTRIBUTING.md).

PR #527, opened 2026-09-07
Support date ranges

fredguth:
`Date` is a single date today, so a range has nowhere to go: the BibLaTeX reader
throws away the START of a `Between`, the CSL-JSON reader panics on a
`date-parts` with two vectors (`panic!("ranges are not supported")`), and
`RenderCsl for citationberg::Date` carries a `// TODO: Date ranges`.

This adds the range, end to end.

## The data

`Date` gains `end: Option<DateEnd>`. `DateEnd` is a separate type rather than a
nested `Date` so that `Date` stays `Copy` and a range cannot nest another range.

The YAML/`FromStr` notation is ISO 8601's time interval, `1989-05/1989-08`;
`Display` and `Serialize` round-trip it. BibLaTeX's `date = {1989-05/1989-08}`
now keeps both ends, and a CSL-JSON `date-parts` with two vectors becomes a
range instead of a panic.

## The rendering

Per CSL 1.0.2 ("Date-part"), the range is rendered once, at the position of its
largest differing date-part, as `start-parts range-delimiter end-parts`; the
date-parts larger than that one are equal for both dates and are rendered once,
in place. The delimiter is the `range-delimiter` of the largest differing part,
defaulting to an en dash.

Two details that the test suite pins down:

- Only date-parts the style actually prints can make a range. Two dates that
  differ only in a day a year-only style never renders are one date to the
  reader, not a range.
- The suffix of the start's last rendered part and the prefix of the end's first
  rendered part make way for the delimiter. Styles that carry their separators
  as affixes on `cs:date-part` — rather than as a `delimiter` on `cs:date` —
  then render `May/August 1989` and not `May /August 1989`. A local test,
  `tests/local/date_RangeAffixes.txt`, covers this case, since the upstream
  suite only exercises the `delimiter` form.

## The tests

`tests/citeproc.rs` no longer skips cases whose input contains a date range.
Eight of them pass now and are added to `citeproc-pass.txt`:
`date_RangeDelimiter`, `date_LocalizedNumericYearRange` and the six
`date_TextForm*Range`. Still failing, and deliberately left out of the passing
file: the open-ended range, the season ranges and the two
`sort_DateVariableRange*` cases (sorting does not look at the end yet).

An open-ended range is the one case that regressed from "panic" to something
visible, so it is handled explicitly: CSL-JSON marks it with an end year of
zero, and since there is no representation for it here yet, the end is dropped
and only the start is rendered — rather than printing `1 B.C.` as the end.

`cargo test --all-features` is green; `cargo fmt` and `cargo clippy
--all-features --all-targets` are clean.


laurmaedje:
https://github.com/typst/hayagriva/pull/523#issuecomment-5569708065

Just opening more AI PRs without any comments after I closed the first one is a little brazen.

If we want to vibecode on Hayagriva, we could do so ourselves.

PR #528, opened 2026-09-08
Fix two container lookups: conference parents and legislation

fredguth:
Two independent fixes in `Entry::get_container`, both found while rendering
Brazilian ABNT NBR 6023:2025 references through a CSL style. Each is one arm of
the same match, so they are sent together.

### 1. Prefer a real container over a conference parent

`get_container` ranked `Conference` above `Periodical`, `Newspaper`, `Blog`,
`Reference` and `Web` for an `Article`. The ranking exists so that an
`@inproceedings` — which yields an Article with both a `Proceedings` and a
`Conference` parent — takes the `Proceedings`, the one that carries the
container's title.

But it also fires when the sibling is a real container. An event published as an
issue of a journal (an Article with a `Periodical` and a `Conference` parent)
lost the journal's title, and its `volume` with it, to the event's name.

A conference is an event, not a container: it stands in as one only when there
is no real container beside it, so it now comes last. `@inproceedings` is
unaffected — `Proceedings` still outranks it.

### 2. Give legislation a container

`get_container` had no arm for `Legislation`, so a statute published in an
official gazette had no container at all: the gazette's title and volume never
reached the style, and only `issue` survived, because it does not go through
`get_container`. A statute is published in a gazette — a newspaper or a
periodical — and reprinted in collections of law, so those are its containers.

---

`cargo test`, `cargo fmt --check` and `cargo clippy --all-targets` are clean.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -804,6 +804,18 @@ impl Entry {
             EntryType::Report => {
                 retrieve_container(&[EntryType::Book, EntryType::Anthology])
             }
+            // A statute is published in an official gazette — a newspaper or a
+            // periodical — and reprinted in collections of law. Without this
+            // arm a `Legislation` had no container at all, so the gazette's
+            // title and volume never reached the style.
+            EntryType::Legislation => retrieve_container(&[
+                EntryType::Newspaper,
+                EntryType::Periodical,
+                EntryType::Anthology,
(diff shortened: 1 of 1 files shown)


laurmaedje:
https://github.com/typst/hayagriva/pull/523#issuecomment-5569708065

